PR #15, opened 2018-11-05
Create a new qpc command for insights upload (#14)

### LINKED ISSUE #14: A new qpc commands for insights upload
- [ ] Define mime type for QPC report data
- [ ] Add code to  QPC to check for insights client installation
- [ ] Retrieve QPC report JSON and compress to tar.gz
- [ ] Add new command to upload file to insights `qpc insights upload --report`

myersCody:
Closes #14


qpc/insights/upload.py
```diff
@@ -0,0 +1,105 @@
...
+
+    SUBCOMMAND = insights.SUBCOMMAND
+    ACTION = insights.UPLOAD
+
+    def __init__(self, subparsers):
+        """Create command."""
+        CliCommand.__init__(self, self.SUBCOMMAND, self.ACTION,
+                            subparsers.add_parser(self.ACTION), GET,
+                            insights.REPORT_URI, [codes.ok])
+        self.parser.add_argument('--report', dest='report_id',
```
abaiken: In our other commands relating to reports, especially downloading (qpc report details/deployments), we can retrieve the report by providing one of two options, either a report id like you do here, or a scan job id. I feel like we should add the option for uploading a report based on scan job to be consistent with our other commands. I don't think it's too much extra to add. What are your thoughts?

abaiken:
We should also document this command inside the `man.rst`

qpc/insights/commands.py
```diff
@@ -0,0 +1,15 @@
+#
+# Copyright (c) 2017 Red Hat, Inc.
```
kholdaway: 2017-2018

qpc/insights/upload.py
kholdaway: That's probably a good idea.  I personally like report id, but consistency is important.  Let's add scan job id.

qpc/insights/upload.py
```diff
@@ -0,0 +1,105 @@
...
+                                 required=True)
+        self.parser.add_argument('--test', dest='test', action='store_true')
+        self.tmp_tar_name = '/tmp/test.tar.gz'
```
kholdaway: Let's make this name random with a timestamp (current millis since epoc)

qpc/insights/upload.py
kholdaway: There is no help messages for the args.

myersCody:
- [ ] Test with old insights client version that doesn't have the upload command on RHEL 6

qpc/insights/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+#
+"""Constants for the Insights commands."""
+
+SUBCOMMAND = 'insights'
+UPLOAD = 'upload'
+
+REPORT_URI = '/api/v1/reports/'
+DEPLOYMENTS_PATH_SUFFIX = '/deployments/'
+# Note the the insights version check does not hangle if versions equal
```
kholdaway: I can hangle this.

qpc/insights/upload.py
```diff
@@ -0,0 +1,166 @@
...
+                            insights.REPORT_URI, [codes.ok])
+        id_group = self.parser.add_mutually_exclusive_group(required=True)
+        id_group.add_argument('--report', dest='report_id',
+                              metavar='REPORT_ID',
+                              help=_(messages.INSIGHTS_REPORT_ID_HELP))
+        id_group.add_argument('--scan-job', dest='scan_job_id',
+                              metavar='SCAN_JOB_ID',
+                              help=_(messages.INSIGHTS_SCAN_JOB_ID_HELP))
+
+        self.parser.add_argument('--test', dest='test', action='store_true')
```
kholdaway: Let's hide --test (see https://stackoverflow.com/questions/11114589/creating-hidden-arguments-with-python-argparse)

qpc/insights/upload.py
```diff
@@ -0,0 +1,171 @@
...
+                            insights.REPORT_URI, [codes.ok])
+        id_group = self.parser.add_mutually_exclusive_group(required=True)
+        id_group.add_argument('--report', dest='report_id',
+                              metavar='REPORT_ID',
+                              help=_(messages.INSIGHTS_REPORT_ID_HELP))
+        id_group.add_argument('--scan-job', dest='scan_job_id',
+                              metavar='SCAN_JOB_ID',
+                              help=_(messages.INSIGHTS_SCAN_JOB_ID_HELP))
+
+        self.parser.add_argument('--test', dest='test', action='store_true',
```
kholdaway: Let's rename --test to --dev.  This is really a development mode not a test.

qpc/insights/upload.py
```diff
@@ -0,0 +1,171 @@
...
+            else:
+                print(_(messages.REPORT_SJ_DOES_NOT_EXIST %
+                        self.args.scan_job_id))
+                sys.exit(1)
+            self.report_source = 'scan job id(%s).' % self.args.scan_job_id
+        else:
+            self.report_id = self.args.report_id
+            self.report_source = 'id (%s).' % self.args.report_id
+        if self.args.test:
+            self.insights_command = InsightsCommands(test=True)
```
kholdaway: rename to `dev` instead of `test`

qpc/insights/upload.py
```diff
@@ -0,0 +1,171 @@
...
+        else:
+            self.insights_command = InsightsCommands()
+        self._check_insights_install()
+        self._check_insights_version()
+        print(_(messages.INSIGHTS_IS_VERIFIED))
+        try:
+            validate_write_file(self.tmp_tar_name, 'tmp_tar_name')
+        except ValueError as error:
+            print(error)
```
kholdaway: let's try and construct a message to help guide the user here.

qpc/insights/utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class InsightsCommands():
+    """Creates insights-client commands for qpc client."""
+
+    def __init__(self, test=False):
+        """Set class variables.
+
+        Args:
+            test (bool): used to test insights commands locally.
```
kholdaway: switch to dev.

qpc/insights/utils.py
```diff
@@ -0,0 +1,120 @@
...
+from __future__ import print_function
+
+# pylint: disable=no-name-in-module,import-error
+from distutils.version import LooseVersion
+
+
+class InsightsCommands():
+    """Creates insights-client commands for qpc client."""
+
+    def __init__(self, test=False):
```
kholdaway: Switch to dev

qpc/insights/utils.py
```diff
@@ -0,0 +1,120 @@
...
+        """Set class variables.
+
+        Args:
+            test (bool): used to test insights commands locally.
+        """
+        self.test = test
+        # mime type for QPC report data for uploading to insights
+        self.content_type = 'application/vnd.redhat.qpc.test+tgz'
```
kholdaway: Based on this documentation, 
```
^application/vnd\.redhat\.([a-z0-9-]+)\.([a-z0-9-]+)\+(tgz|zip)$, where the first matching group is a unique name for an Insights Platform application, the second matching group identifies the payload type for that application, and the third matching group identifies whether the payload is a gzipped tarball or zip file. Any other Content-Type header values will be met with a 415 Unsupported Media Type response. These supported types should be discoverable with a GET request.
```

Should the content type be `application/vnd.redhat.qpc.deployments.tgz`?

qpc/insights/utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def build_base(self):
+        """Will create a base used for creating insights commands."""
+        if self.test:
```
kholdaway: change it dev

qpc/messages.py
```diff
@@ -200,9 +200,9 @@
     'Scan Job %s does not exist.'
 REPORT_SJS_DO_NOT_EXIST = 'The following scan jobs do not exist: %s.'
 REPORT_NO_DEPLOYMENTS_REPORT_FOR_SJ = \
-    'No report summary available for scan job %s.'
+    'No deployments report available for scan job %s.'
 REPORT_NO_DEPLOYMENTS_REPORT_FOR_REPORT_ID = \
-    'No report summary available for report id %s.'
+    'No deployments report available for report id %s.'
```
kholdaway: `Report %s does not exist.`

qpc/insights/utils.py
myersCody: I was not certain if we had to have the insights team build us an endpoint or not? Honestly, I am not exactly sure how they use content type/mime type.

qpc/insights/upload.py
myersCody: Do you want me to do this for the deployments & details report too?

myersCody: `def validate_write_file(filename, param_name):
    """Write content to a file.

    :param filename: the filename to write
    :param param_name: the parameter name that provided
    the filename
    :raises: ValueError for validation errors
    """
    input_path = os.path.expanduser(os.path.expandvars(filename))
    if not input_path:
        raise ValueError(
            t(messages.REPORT_OUTPUT_CANNOT_BE_EMPTY % param_name))
    if os.path.isdir(input_path):
        raise ValueError(
            t(messages.REPORT_OUTPUT_IS_A_DIRECTORY %
              (param_name, input_path)))
    directory = os.path.dirname(input_path)
    if directory and not os.path.exists(directory):
        raise ValueError(
            t(messages.REPORT_DIRECTORY_DOES_NOT_EXIST % directory))`

I believe we overwrite the ValueError inside of `validate_write_file` to help guide the user.  We use the same format for details & deployments reports.

qpc/insights/utils.py
```diff
@@ -0,0 +1,121 @@
...
+            dev (bool): used to build dev insights commands for local use.
+        """
+        self.dev = dev
+        # mime type for QPC report data for uploading to insights
+        self.content_type = 'application/vnd.redhat.qpc.deployments+tgz'
+
+    def build_base(self):
+        """Will create a base used for creating insights commands."""
+        if self.dev:
+            self.content_type = 'application/vnd.redhat.qpc.test+tgz'
```
myersCody: I added this incase we wanted a separate endpoint for development.

qpc/insights/upload.py
kholdaway: Can you post the output?  I want to make sure the message is clean and not the printout of an object.  Let's not use the same format if it is not ideal.  I would lean towards changing that also.

qpc/insights/utils.py
kholdaway: We should use that to specify what we are sending.  In this case, we are sending the deployments report.

PR #35, opened 2018-12-07
CLI side verification 

### LINKED ISSUE #32: User Story: Add QPC client side deployment report validation for insights upload command
## Feature description:
qpc should warn users that a deployment report contains system fingerprints that do not have canonical facts.  The subscription insights service will reject invalid reports so we should validate locally.

### Validation points:
- If every system fingerprint is missing all insights host canonical facts.  The report is not valid and has no value.  Log error and exit without sending the report.
- If the above is not true and some systems are missing all insights host canonical facts, then log the system fingerprint with a warning.  Indicate that this particular system will not be included in the deployment report.  The QPC CLI should remove invalid system fingerprints before sending the report.

abaiken:
Closes #32


qpc/insights/upload.py
```diff
@@ -147,16 +168,54 @@ def _validate_args(self):
...
         print(_(messages.INSIGHTS_RETRIEVE_REPORT % self.report_id))
 
+    def verify_report_details(self):
+        """
+        Verify that the report contents are a valid deployments report.
+
+        :returns boolean regarding report validity
+        """
+        # validate required keys
+        report_contents = extract_json_from_tarfile(self.tmp_tar_name)
```
myersCody: I think that you can validate from memory before the tar file is created.

qpc/insights/upload.py
```diff
@@ -30,13 +30,34 @@
...
+        found_facts = False
+        for fact in canonical_facts:
+            if fingerprint.get(fact):
+                found_facts = True
+                break
+        if found_facts:
+            verified_fingerprints.append(fingerprint)
+        else:
+            print(_(messages.INSIGHTS_REPORT_INVALID_FP %
+                    (report_id, fingerprint.pop('metadata', None))))
```
myersCody: If there is no metadata that means that a None will be returned for the fingerprint, even though there may be some other keys in there. There should always be metadata so it may not be an issue.

qpc/insights/upload.py
abaiken: Great catch! Fixed

qpc/insights/upload.py
```diff
@@ -30,13 +30,35 @@
...
-from qpc.utils import (validate_write_file,
+from qpc.utils import (extract_json_from_tar,
+                       validate_write_file,
                        write_file)
 
 # pylint:disable=no-member
 from requests import codes
 
 
+def verify_report_fingerprints(fingerprints, report_id):
```
kholdaway: Let's have this return the good/bad fingerprints as a tuple.

qpc/insights/upload.py
```diff
@@ -30,13 +30,35 @@
...
+        found_facts = False
+        for fact in canonical_facts:
+            if fingerprint.get(fact):
+                found_facts = True
+                break
+        if found_facts:
+            verified_fingerprints.append(fingerprint)
+        else:
+            fingerprint.pop('metadata', None)
+            print(_(messages.INSIGHTS_REPORT_INVALID_FP %
```
kholdaway: Let's not log the entire fingerprint (opposite of yupana).  Here the user is close to the data.  The fingerprint's name has more meaning to the end user.  Additionally, original source names mean something.  Let's create a log entry that is one line per fingerprint with meaningful information from meta-data.

qpc/insights/upload.py
```diff
@@ -147,13 +169,51 @@ def _validate_args(self):
...
+        report_type = report_contents.get('report_type')
+        if report_type != 'deployments':
+            print(_(messages.INSIGHTS_INVALID_REPORT_TYPE % report_id))
+            return False
+
+        # validate fingerprints contain canonical facts
+        fingerprints = report_contents.get('system_fingerprints')
+        verified_fingerprints = verify_report_fingerprints(fingerprints, report_id)
+        if not verified_fingerprints:
+            print(_(messages.INSIGHTS_REPORT_NO_VALID_FP % report_id))
```
kholdaway: If there are some invalid fingerprints, we should still send.  This log message should be changed.  If there are no valid fingerprints, then the report is invalid.

qpc/insights/upload.py
```diff
@@ -147,13 +169,51 @@ def _validate_args(self):
...
+        # validate report type
+        report_type = report_contents.get('report_type')
+        if report_type != 'deployments':
+            print(_(messages.INSIGHTS_INVALID_REPORT_TYPE % report_id))
+            return False
+
+        # validate fingerprints contain canonical facts
+        fingerprints = report_contents.get('system_fingerprints')
+        verified_fingerprints = verify_report_fingerprints(fingerprints, report_id)
+        if not verified_fingerprints:
```
kholdaway: Let's print some summary stats here if some fingerprints are invalid.  For example, x invalid fingerprints.  We can still send the report (for now), but should warn that some of the fingerprints are going to be rejected.

qpc/insights/upload.py
```diff
@@ -147,13 +183,51 @@ def _validate_args(self):
...
+        # validate required keys
+        report_contents = extract_json_from_tar(self.response.content,
+                                                print_pretty=False)
+        required_keys = ['report_platform_id',
+                         'report_id',
+                         'report_version',
+                         'report_type',
+                         'system_fingerprints']
+        for key in required_keys:
+            present = report_contents.get(key)
```
kholdaway: Let's gather all the keys that are missing and generate a message that lists all of them.

PR #75, opened 2019-04-03
Changed build process to use package name as a variable.

### LINKED ISSUE #74: Modify egg building process to use a package name variable
## Feature description:
We want to control the package name through a global variable instead of hardcoded to make the egg building process easier for downstream.

myersCody:
Closes #74

### DIFF
--- a/qpc/utils.py
+++ b/qpc/utils.py
@@ -408,12 +408,3 @@ def create_tar_buffer(files_data):
             tar_file.addfile(tarinfo=info, fileobj=file_buffer)
     tar_buffer.seek(0)
     return tar_buffer.getvalue()
-
-
-def get_settings():
-    """Collect the cli version from settings json."""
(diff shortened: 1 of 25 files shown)


qpc/utils.py
```diff
@@ -408,3 +408,12 @@ def create_tar_buffer(files_data):
             tar_file.addfile(tarinfo=info, fileobj=file_buffer)
     tar_buffer.seek(0)
     return tar_buffer.getvalue()
+
+
+def get_settings():
```
kholdaway: Let's not stick this in util as a function.  So I think the settings should be it's own module (e.g. release_settings.py).  I think it should load the json on startup.  Additionally, have accessor methods.  Let's chat tomorrow, but I think it will be a little different from this.

kholdaway:
Looks good when codecov passes.

PR #78, opened 2019-04-05
Increase version of CLI to 1.0.0

kholdaway:


myersCody:
Should we change issue #71 to say 1.0.0?

PR #80, opened 2019-04-05
Add flag to allow CLI to disable authentication

kholdaway:


qpc/utils.py
```diff
@@ -151,11 +164,19 @@ def read_server_config():
...
+
         if not isinstance(use_http, bool):
             log.error('Server config %s has invalid value for use_http %s',
                       QPC_SERVER_CONFIG, use_http)
             return None
 
+        if not isinstance(require_token, bool):
+            log.error('Server config %s has invalid value for require_token %s',
+                      QPC_SERVER_CONFIG, require_token)
+            return None
```
abaiken: instead of returning `None` here could we just default it to True and still return the dictionary? that way we don't have to check for `None` in `read_require_auth`

kholdaway: it returns true in that case.

abaiken: I get that, its just kind of a weird way around it. Like we could check for if the `require_token` is `None` or is not an instance of a bool & return `True` in both cases instead of `True` in one, and `None` in the other. I get that is how the other values were done but it seems odd.

kholdaway: This is the pattern of that file.  We can open an issue to re-write it.  The others are similar.  Basically, if there is invalid JSON, it returns None for the entire config file.

PR #102, opened 2019-05-24
Update the text in make help

myersCody:


Makefile
```diff
@@ -70,5 +65,5 @@ insights-client:
 
 
 insights-clean:
-	sudo rm -rf /etc/insights-client/* 
+	sudo rm -rf /etc/insights-client/*
```
abaiken: we should add `all` and `build` to the help as well

myersCody: Good point 👍

Makefile
```diff
@@ -70,5 +67,5 @@ insights-client:
...
```
kholdaway: This change just seems like whitespace.  Just revert that.

PR #146, opened 2019-11-26
Add a command to validate a reports authenticity

abaiken:
Added `qpc report validate` command.
To test: 
1. Make sure you are using quipucords branch `issues/2018` 
2. Run a scan and download a tar.gz file containing all of the reports 
3. Open the tar.gz and get the hash associated with a report. Use the `qpc report validate` command to pass in a report and its associated hash to see if it is authentic


abaiken:
Closing this in favor of doing validating through unix/linux commands.

PR #187, opened 2022-06-30
Fix/protect config files

bruno-fs:
- Prevent user config files from being modified during test execution.
- Replace unittest as test runner with pytest


gtanzillo (approved):
Looks good 👍

nicolearagao (approved):
Great work!

PR #189, opened 2022-07-14
Add HTTP client for console.redhat.com

nicolearagao:
Create class InsightsClient which will be responsible for handling
requests to C.RH.C.

This is part of the implementation of Discovery-131 feature.

### DIFF
--- a/qpc/insights/test_insights_http.py
+++ b/qpc/insights/test_insights_http.py
@@ -1,0 +1,92 @@
+# Copyright (C) 2022  Red Hat, Inc.
+
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+"""Tests for the HTTP Client for console.redhat.com."""
+from unittest import mock
+
+import pytest
+
+from qpc.insights.http import InsightsClient
+
+
+@mock.patch("requests.Session.request")
+@pytest.mark.parametrize(
+    "base_url, method, url, result",
+    [
+        (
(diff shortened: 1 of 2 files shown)


qpc/insights/test_insights_http.py
```diff
@@ -0,0 +1,92 @@
...
+from unittest import mock
+
+import pytest
+
+from qpc.insights.http import InsightsClient
+
+
+@mock.patch("requests.Session.request")
+@pytest.mark.parametrize(
+    "base_url, method, url, result",
```
bruno-fs, line 19: I'd rename `result` to `final_url`

qpc/insights/test_insights_http.py
```diff
@@ -0,0 +1,92 @@
...
+            "https://testconsole.redhat.com/v1/path/to/resource",
+        ),
+    ],
+)
+def test_request_joins_url_with_base_url(
+    request_session_mocker, base_url, method, url, result
+):
+    """Test if urljoin is performed successfully."""
+    insights_client_session = InsightsClient(
+        base_url=base_url, auth=("username", "pass")
```
bruno-fs, line 46: In both tests I don't see auth being tested - maybe its only used after the mocked point.

That's fine, but since they are not in use I'd remove the auth keyword from both tests and add a simple
unit test like this

```
def test_auth():
    client = InsigthsClient(auth=("foo", "bar"))
    assert client.auth == ("foo", "bar")
```

We don't need more than that because:
1. the integration test on camayoc will take care of the rest.
2. we trust requests.Session internal API stability XD

qpc/insights/test_insights_http.py
gtanzillo: or maybe `full_url`

bruno-fs: even better

nicolearagao: done

gtanzillo (approved):
Looks good 👍

bruno-fs (approved):
LGTM :+1:

PR #191, opened 2022-07-20
Fix a bug on json_dir treatment

bruno-fs:
Example of the problem here:
https://github.com/quipucords/qpc/runs/7438213333?check_suite_focus=true#step:5:242

if json_dir is not a list it will pick the first letter of the path, which would be the root dir (`/`).


nicolearagao (approved):
Awesome work

gtanzillo (approved):
LGTM

PR #192, opened 2022-07-22
Insights login password encryption

nicolearagao:
-Create encrypt/decrypt functions using the fernet algorithm, before
savig it to insights-encryption file. The file permissions are set to be
read only by the user. Decryption won't occur if permissions are too
broad.
-Create class of QPC base exception, and insights exceptions.

This is part of the implementatio of Discovery-131 feature


qpc/utils.py
```diff
@@ -520,3 +529,47 @@ def check_extension(extension, path):
...
+    write_encryption_key_if_non_existent()
+    key = load_encryption_key()
+
+    encryption_algorithm = Fernet(key)
+
+    encrypted_password = encryption_algorithm.encrypt(password.encode())
+    return encrypted_password.decode()
+
+
+def decrypt_password(password):
```
nicolearagao, line 569: @bruno-fs Hey, you asked me to ass an exception here in case anything goes wrong with decryption, but I believe we should add to the function that calls read_insights_login_config instead. This will happen as part of the publish implementation. 
What do you and @gtanzillo think?

bruno-fs: I'll start by the end: yes, the place to properly treat the exception is the publish command.

but it shouldn't deal with cryptography exceptions, only QPCErrors. the only functios that know about python cryptography are the encryption/decryption related ones - and in my view is their responsibility to "convert" expected 3rd party exceptions to the ones qpc entrypoints expects

gtanzillo: I'm 👍 with what @bruno-fs said ^^. Agree that the exception handling for encrypt/decrypt should be at the place where those operations occur. They can always be handled and then bubbled up the call stack if it's necessary for the calles to also handle them

nicolearagao: I understand and agree with both of you. I'll push a fixup with those changes and after the publish pr and this one get merged I'll add a QPCError whenever read_insights_login_config is called.

bruno-fs: good call

bruno-fs (approved):
LGTM :+1:

gtanzillo (approved):
LGTM too 👍

PR #194, opened 2022-07-27
Remove insights upload from cli commands

nicolearagao:
Removal is neccessary to simplify things for the end user, who might confuse the upload and publish subcommands.
Code was not completely removed, in case we need it in the future.


nicolearagao:
Captura de tela de 2022-07-27 18-00-29
@bruno-fs @gtanzillo @ruda

gtanzillo (approved):
Thanks @nicolearagao 👍

bruno-fs (approved):
:+1:

PR #195, opened 2022-07-28
[DISCOVERY-166] validate insights input file

bruno-fs:
- Validate insights report tarball
- Log ingress response

### DIFF
--- a/qpc/insights/publish.py
+++ b/qpc/insights/publish.py
@@ -11,6 +11,7 @@
 
 import os
 import sys
+import tarfile
 from logging import getLogger
 from pathlib import Path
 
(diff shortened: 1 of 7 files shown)


qpc/utils.py
```diff
@@ -340,18 +340,14 @@ def setup_logging(verbosity):
     else:
         log_level = logging.DEBUG
 
+    log_fmt = "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
```
nicolearagao, line 343: I know that the log changes are mostly for the end user, but do you think we could benefit from adding "%(funcName)s:%(lineno)d" to formatter whenever we are on debug verbosity level?

bruno-fs: makes sense

qpc/utils.py
```diff
@@ -340,18 +340,14 @@ def setup_logging(verbosity):
...
-    # but we only adjust the log level for the 'qpc' logger.
-    log.setLevel(log_level)
-    # the StreamHandler sends warnings and above to stdout, but
-    # only for messages going to the 'qpc' logger, i.e. QPC
-    # output.
-    stderr_handler = logging.StreamHandler()
-    stderr_handler.setLevel(logging.ERROR)
-    log.addHandler(stderr_handler)
+    logging.basicConfig(filename=QPC_LOG, format=log_fmt, level=log_level)
+    stream_handler = logging.StreamHandler()
```
nicolearagao, line 347: What do you think about making it a separate function?

```
def get_console_handler(log_fmt, log_level):
       setup ...
        return stream_handler
```

bruno-fs: Mmm makes sense but I think this is small enough to justify not adding the "bureaucracy" of an extra function XD

qpc/utils.py
bruno-fs: any thoughts on this https://github.com/quipucords/qpc/pull/195/commits/16f36451f4cdfa1f2e34f477e11b314b5d29edcd?

qpc/insights/publish.py
```diff
@@ -61,12 +62,63 @@ def _validate_insights_report_name(self):
...
+        except tarfile.ReadError as err:
+            log.exception(_(messages.INSIGHTS_REPORT_CONTENT_UNEXPECTED))
+            raise SystemExit(1) from err
+        return filenames
+
+    def _is_top_folder(self, top_folder):
+        expected_top_folder_parent = Path(".")
+        return top_folder.parent == expected_top_folder_parent
+
+    def _separate_top_folder(self, filenames):
```
nicolearagao, line 95: nice!

qpc/insights/tests_insights_publish.py
```diff
@@ -148,7 +196,7 @@ def test_insights_publish_successful(  # pylint: disable=too-many-arguments
             "insights",
             "publish",
             "--input-file",
-            payload_file.name,
+            str(payload_file),
```
nicolearagao, line 203: why not use name attribute ?

qpc/utils.py
nicolearagao: Fair enough =)

qpc/insights/tests_insights_publish.py
bruno-fs: it wont give the full path for the file. I didn't dig on the explanation, but the file produced by `tmp_path` fixture don't behave the same way as `NamedTemporaryFile`

bruno-fs: Now that I thought "out loud" about it makes sense to behave differently. `tmp_path` is a `pathlib.Path` object in the end :thinking:

nicolearagao: got it.

bruno-fs: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name

nicolearagao (approved):
LGTM 👍🏼

qpc/insights/publish.py
```diff
@@ -61,12 +62,63 @@ def _validate_insights_report_name(self):
             log.info(_(messages.INSIGHTS_LOCAL_REPORT_NOT_TAR_GZ))
             sys.exit(1)
 
+    def _validate_insights_report_content(self):
+        """Check if report tarball contains the expected json files."""
+        filenames = self._get_filenames()
+
+        if len(filenames) < 2:
+            log.error(_(messages.INSIGHTS_REPORT_CONTENT_MIN_NUMBER))
```
gtanzillo, line 70: Should all of these error messages related to validation failures be prefixed with something like `Payload file is not valid, due to: <specific error message>`?

bruno-fs: I think it's a good idea! I'm thinking on a adding this prefix directly on directly on the messages module (https://github.com/quipucords/qpc/blob/16f36451f4cdfa1f2e34f477e11b314b5d29edcd/qpc/messages.py#L362-L370).
I think I'd have to adapt a little bit the messages to accommodate the prefix :thinking:  
What do you think?

bruno-fs: (the linked snippet contains only the new messages)

gtanzillo: Adding the prefix directly to the messages is plenty good enough for this :+1

bruno-fs: done!

gtanzillo (approved):
Looks great! Thanks @bruno-fs

PR #198, opened 2022-08-03
[DISCOVERY-168] Drop support for python 3.6 and pin dependencies

bruno-fs:
- Drop support for python 3.6
- Adopt poetry to manage dependencies (since we don't need cachito grab dependencies we don't need pip-tools)
- Remove unused/unnecessary dev dependencies


bruno-fs:
please ignore the multiple commits when reviewing - just see the final changes on the files. I'll squash them all in the end

gtanzillo (approved):
LGTM 👍

nicolearagao (approved):
LGTM 👍🏼

PR #199, opened 2022-08-03
[DISCOVERY-169] - Make Log handlers more friendly to camayoc

bruno-fs:
Fix DISCOVERY-169

### DIFF
--- a/qpc/utils.py
+++ b/qpc/utils.py
@@ -352,7 +352,11 @@ def setup_logging(verbosity):
     # those not coming from qpc, will go to the log file
     logging.basicConfig(filename=QPC_LOG, format=log_fmt, level=log_level)
     stream_handler = logging.StreamHandler()
-    stream_handler.setFormatter(logging.Formatter(log_fmt))
+    if log_level == logging.DEBUG:
+        # changing log format was breaking camayoc tests. let's add this extra logging
+        # information only when the user actually requests for more logs.
+        # (at least until we add an option controlling the log format)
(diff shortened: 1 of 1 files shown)


bruno-fs:
tests will break until #198 gets merged

qpc/utils.py
```diff
@@ -352,7 +352,11 @@ def setup_logging(verbosity):
     # those not coming from qpc, will go to the log file
     logging.basicConfig(filename=QPC_LOG, format=log_fmt, level=log_level)
     stream_handler = logging.StreamHandler()
-    stream_handler.setFormatter(logging.Formatter(log_fmt))
+    if log_level == logging.DEBUG:
+        # changing log format was breaking camayoc tests. let's add this extra logging
+        # information only when the user actually requests for more logs.
+        # (at least until we add an option controlling the log format)
```
nicolearagao, line 358: Will we add this info to documentation as well, so the user knows he has this option?

nicolearagao (approved):
LGTM 👍🏼

gtanzillo (approved):
LGTM 👍

qpc/utils.py
bruno-fs: We can point the user can pass -v or -vv for more verbosity on any command if its not already mentioned.

What do you think @justinkreft ?

bruno-fs: But honestly I expect any cli user to run `-h`, `--help` or `man <COMMAND>` when looking for instructions.

nicolearagao: Fair point

PR #204, opened 2022-09-08
Feature/combine insights download publish

nicolearagao:
Add new argument --report to InsightsPublishCommand which downloads the
 insights report from qpc server and publishes it to insights.
 
Implementation of DISCOVERY-170

A few observations regarding linting:
- publish.py line 183 disable=consider-using-with(I don't think a context manager is necessary in this situation since we make sure to delete the file after it was published)

### DIFF
--- a/qpc/insights/tests_insights_publish.py
+++ b/qpc/insights/tests_insights_publish.py
@@ -18,7 +18,8 @@ import pytest
 
 from qpc import messages
 from qpc.cli import CLI
-from qpc.insights import INGRESS_REPORT_URI
+from qpc.insights import INGRESS_REPORT_URI, INSIGHTS_PATH_SUFFIX, REPORT_URI
+from qpc.insights.publish import InsightsPublishCommand
 from qpc.utils import write_server_config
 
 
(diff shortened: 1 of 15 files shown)


bruno-fs (changes requested):
It's good but has some unnecessary extra code.

qpc/insights/tests_insights_publish.py
```diff
@@ -278,3 +300,104 @@ def test_invalid_payload_file(self, payload, caplog, log_message):
...
+            "/bin/qpc",
+            "insights",
+            "publish",
+            "--report",
+            "1",
+        ]
+        CLI().main()
+
+        mock_validate_report_name.assert_called_with(str(payload_file))
+        mock_validate_report_content.assert_called_with(str(payload_file))
```
bruno-fs, line 401: is this working? :open_mouth: 

it should be throwing a NameError :scream:

bruno-fs (approved):

gtanzillo (approved):
Looks Great! Thanks @nicolearagao

PR #205, opened 2022-10-11
Feature/add openshift credential

nicolearagao:
Add openshift credential to CLI.

This pr is cli implementation of task [DISCOVERY-185].

### DIFF
--- a/qpc/cred/add.py
+++ b/qpc/cred/add.py
@@ -47,14 +47,16 @@ class CredAddCommand(CliCommand):
         self.parser.add_argument('--type', dest='type',
                                  choices=[credential.NETWORK_CRED_TYPE,
                                           credential.VCENTER_CRED_TYPE,
-                                          credential.SATELLITE_CRED_TYPE],
+                                          credential.SATELLITE_CRED_TYPE,
+                                          credential.OPENSHIFT_CRED_TYPE],
                                  metavar='TYPE',
                                  help=_(messages.CRED_TYPE_HELP),
+                                 type=str.lower,
                                  required=True)
         self.parser.add_argument('--username', dest='username',
                                  metavar='USERNAME',
                                  help=_(messages.CRED_USER_HELP),
-                                 required=True)
+                                 required=False)
         group = self.parser.add_mutually_exclusive_group(required=True)
         group.add_argument('--password', dest='password',
                            action='store_true',
@@ -62,6 +64,9 @@ class CredAddCommand(CliCommand):
         group.add_argument('--sshkeyfile', dest='filename',
                            metavar='FILENAME',
                            help=_(messages.CRED_SSH_HELP))
+        group.add_argument("--token", dest="token",
+                           action="store_true",
+                           help=_(messages.CRED_TOKEN_HELP))
         self.parser.add_argument('--sshpassphrase', dest='ssh_passphrase',
                                  action='store_true',
                                  help=_(messages.CRED_SSH_PSPH_HELP))
--- a/qpc/cred/add.py
+++ b/qpc/cred/add.py
@@ -41,45 +41,78 @@ class CredAddCommand(CliCommand):
                             subparsers.add_parser(self.ACTION), POST,
                             credential.CREDENTIAL_URI, [codes.created])
 
-        self.parser.add_argument('--name', dest='name', metavar='NAME',
-                                 help=_(messages.CRED_NAME_HELP),
-                                 required=True)
-        self.parser.add_argument('--type', dest='type',
-                                 choices=[credential.NETWORK_CRED_TYPE,
-                                          credential.VCENTER_CRED_TYPE,
-                                          credential.SATELLITE_CRED_TYPE,
-                                          credential.OPENSHIFT_CRED_TYPE],
-                                 metavar='TYPE',
-                                 help=_(messages.CRED_TYPE_HELP),
-                                 type=str.lower,
-                                 required=True)
-        self.parser.add_argument('--username', dest='username',
-                                 metavar='USERNAME',
-                                 help=_(messages.CRED_USER_HELP),
-                                 required=False)
+        self.parser.add_argument(
+            "--name",
+            dest="name",
+            metavar="NAME",
+            help=_(messages.CRED_NAME_HELP),
+            required=True,
+        )
+        self.parser.add_argument(
+            "--type",
+            dest="type",
+            choices=[
+                credential.NETWORK_CRED_TYPE,
+                credential.VCENTER_CRED_TYPE,
+                credential.SATELLITE_CRED_TYPE,
+                credential.OPENSHIFT_CRED_TYPE,
+            ],
+            metavar="TYPE",
+            help=_(messages.CRED_TYPE_HELP),
+            type=str.lower,
+            required=True,
+        )
+        self.parser.add_argument(
+            "--username",
+            dest="username",
+            metavar="USERNAME",
+            help=_(messages.CRED_USER_HELP),
+            required=False,
+        )
         group = self.parser.add_mutually_exclusive_group(required=True)
-        group.add_argument('--password', dest='password',
-                           action='store_true',
-                           help=_(messages.CRED_PWD_HELP))
-        group.add_argument('--sshkeyfile', dest='filename',
-                           metavar='FILENAME',
-                           help=_(messages.CRED_SSH_HELP))
-        group.add_argument("--token", dest="token",
-                           action="store_true",
-                           help=_(messages.CRED_TOKEN_HELP))
-        self.parser.add_argument('--sshpassphrase', dest='ssh_passphrase',
-                                 action='store_true',
-                                 help=_(messages.CRED_SSH_PSPH_HELP))
-        self.parser.add_argument('--become-method', dest='become_method',
-                                 choices=credential.BECOME_CHOICES,
-                                 metavar='BECOME_METHOD',
-                                 help=_(messages.CRED_BECOME_METHOD_HELP))
-        self.parser.add_argument('--become-user', dest='become_user',
-                                 metavar='BECOME_USER',
-                                 help=_(messages.CRED_BECOME_USER_HELP))
-        self.parser.add_argument('--become-password', dest='become_password',
-                                 action='store_true',
-                                 help=_(messages.CRED_BECOME_PASSWORD_HELP))
+        group.add_argument(
+            "--password",
+            dest="password",
+            action="store_true",
+            help=_(messages.CRED_PWD_HELP),
+        )
+        group.add_argument(
+            "--sshkeyfile",
+            dest="filename",
+            metavar="FILENAME",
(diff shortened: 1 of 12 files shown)


nicolearagao:
@gtanzillo @bruno-fs 
I see that our coverage for qpc/cred/utils.py went down a bit. Currently there are not unit tests for the functions in that module(they are the ones who build the payload). Should I add some unit tests for those as well or you dont think they are necessary?

qpc/cred/utils.py
```diff
@@ -45,6 +45,12 @@ def get_password(args, req_payload, add_none=True):
         req_payload['become_password'] = pass_prompt or None
     elif add_none:
         req_payload['become_password'] = None
+    if "token " in args and args.token:
+        print(_(messages.OPENSHIFT_TOKEN))
+        token_prompt = getpass()
+        req_payload["auth_token"] = token_prompt or None
```
bruno-fs, line 51: Instead of allowing None or "", check, raise an error.

qpc/cred/utils.py
```diff
@@ -45,6 +45,12 @@ def get_password(args, req_payload, add_none=True):
         req_payload['become_password'] = pass_prompt or None
     elif add_none:
         req_payload['become_password'] = None
+    if "token " in args and args.token:
```
bruno-fs, line 48: why this is space is expected ("token ")? and why args.token is not enough?
is this a protection against AttributeError? if so, `if getattr(args, "token", None)` would be enough

qpc/cred/add.py
```diff
@@ -47,21 +47,26 @@ def __init__(self, subparsers):
...
+                                          credential.OPENSHIFT_CRED_TYPE],
                                  metavar='TYPE',
                                  help=_(messages.CRED_TYPE_HELP),
+                                 type=str.lower,
                                  required=True)
         self.parser.add_argument('--username', dest='username',
                                  metavar='USERNAME',
                                  help=_(messages.CRED_USER_HELP),
-                                 required=True)
+                                 required=False)
```
bruno-fs: I wonder if this change won't bring unintended side effects to other credential types :thinking:

qpc/cred/test_openshift_cred_add.py
```diff
@@ -0,0 +1,169 @@
...
+        }
+    )
+
+
+@pytest.fixture
+def openshift_token_input(monkeypatch):
+    """Mock Openshift token return from prompt."""
+    with monkeypatch.context() as openshift_token:
+        openshift_token.setattr("builtins.input", lambda _: "token_from_prompt")
+        yield openshift_token
```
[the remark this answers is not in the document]
bruno-fs: wouldn't be easier just mock `getpass`?

bruno-fs (commented):
Love that we have more tests than code here! =D
Made some small suggestions.

qpc/cred/test_openshift_cred_add.py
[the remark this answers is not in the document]
nicolearagao: Good advice, I will mock getpass instead.

qpc/cred/test_openshift_cred_add.py
```diff
@@ -0,0 +1,169 @@
...
+    """Class for testing OpenShift add credential."""
+
+    @pytest.mark.parametrize(
+        "status_code,err_log_message",
+        [
+            (401, "qpc server login"),
+            (500, "An internal server error occurred."),
+        ],
+    )
+    def test_openshift_cred_add_returning_error(  # pylint: disable=too-many-arguments
```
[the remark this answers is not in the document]
nicolearagao: I modified the funcs names.

qpc/cred/test_openshift_cred_add.py
```diff
@@ -0,0 +1,169 @@
...
+        out, err = capsys.readouterr()
+        assert out == ""
+        assert "one of the arguments --password --sshkeyfile --token is required" in err
+
+    def test_openshift_cred_add_no_name(
+        self,
+        capsys,
+        openshift_token_input,
+    ):
+        """Test openshift cred add no type."""
```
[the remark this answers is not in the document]
nicolearagao: ops

qpc/cred/test_openshift_cred_edit.py
```diff
@@ -0,0 +1,179 @@
...
+        out, err = capsys.readouterr()
+        assert out == ""
+        assert err_log_message in err
+
+    def test_edit_openshift_cred_incorrect_arg(
+        self,
+        capsys,
+        requests_mock,
+    ):
+        """Test openshift edit cred successfully."""
```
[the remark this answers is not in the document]
nicolearagao: ops

qpc/cred/test_openshift_cred_list.py
```diff
@@ -0,0 +1,87 @@
...
+        CLI().main()
+        cred_types_except_ocp = (
+            SATELLITE_CRED_TYPE,
+            VCENTER_CRED_TYPE,
+            NETWORK_CRED_TYPE,
+        )
+        out, err = capsys.readouterr()
+        assert any(cred in out for cred in ("openshift_2", "openshift_1"))
+        assert out.count("cred") == 2
+        assert not any(cred_type in out for cred_type in cred_types_except_ocp)
```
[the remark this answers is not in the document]
nicolearagao: I was mistaken yesterday, out is not a json. It's a string. 
`'[\n    {\n        "cred_type": "openshift",\n        "id": 1,\n        "name": "openshift_1",\n        "token": "********"\n    },\n    {\n        "cred_type": "openshift",\n        "id": 2,\n        "name": "openshift_2",\n        "token": "********"\n    }\n]\n'`
Is there another way to assert the output is what it is expected?

qpc/cred/utils.py
nicolearagao: I wrote a function that will validate the pass_prompt and raise an error if empty or none.

qpc/cred/utils.py
nicolearagao: the space in token was an accident. I will push a fixup with a correction.

qpc/cred/add.py
nicolearagao: Hm. I know that this is dealt with in the server side.  There's no function in the cli that validates the fields for each cred type.

qpc/cred/test_openshift_cred_edit.py
```diff
@@ -0,0 +1,179 @@
...
+        self,
+        capsys,
+        requests_mock,
+        openshift_token_input,
+        status_code,
+        err_log_message,
+    ):
+        """Test openshift edit cred successfully."""
+        url_get = get_server_location() + CREDENTIAL_URI
+        requests_mock.get(url_get, status_code=status_code, json=None)
```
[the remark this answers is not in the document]
nicolearagao: A 400 patch error is covered on def test_edit_incorrect_arg. Do you think we should add any other error? I couldnt think of other scenarios to test besides a 400.

qpc/cred/utils.py
nicolearagao: @bruno-fs IT does need to have both if "token " in args and args.token, otherwise it will raise an attributeError.

bruno-fs: ok. I suggest using getattr, then (see my first comment in this thread)

qpc/cred/add.py
bruno-fs: OK, then!

qpc/cred/test_openshift_cred_edit.py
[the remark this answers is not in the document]
bruno-fs: Oh, OK. At the time I asked it was making a patch call with expected 200 status code

qpc/cred/conftest.py
```diff
@@ -20,7 +20,7 @@ def _setup_server_config_file():
     Since all cli commands require qpc config and qpc login to be executed,
     require_token must be False, to avoid passing login info
     """
-    from qpc.utils import write_server_config
+    from qpc.utils import write_server_config  # pylint: disable=import-outside-toplevel
```
bruno-fs, line 23: because this type of import will be common on conftest.py you may put disable this lint for the whole file

qpc/cred/test_openshift_cred_list.py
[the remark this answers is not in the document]
bruno-fs: you sure this isn't a json? looks like it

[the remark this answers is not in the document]
bruno-fs: yeah... it is.

```json
[
    {
        "cred_type": "openshift",
        "id": 1,
        "name": "openshift_1",
        "token": "********"
    },
    {
        "cred_type": "openshift",
        "id": 2,
        "name": "openshift_2",
        "token": "********"
    }
]
```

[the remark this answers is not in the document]
bruno-fs: you need to import json module and then

```
out_as_dict = json.loads(out)
```
Then you will have it as dict.

qpc/cred/test_openshift_cred_edit.py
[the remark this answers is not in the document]
bruno-fs: replace "edit cred successfully" with "edit credential with incorrect arguments" or something like that.

qpc/cred/utils.py
bruno-fs: its technically impossible to have `token_prompt` "falsy" here. you can drop the "or" part.

bruno-fs (approved):

qpc/cred/test_openshift_cred_edit.py
```diff
@@ -0,0 +1,164 @@
...
+        requests_mock,
+    ):
+        """Test openshift edit cred with incorrect args."""
+        url_get = get_server_location() + CREDENTIAL_URI
+        url_patch = get_server_location() + CREDENTIAL_URI + "1/"
+        results = [
+            {
+                "id": 1,
+                "name": "openshift_cred",
+                "cred_type": OPENSHIFT_CRED_TYPE,
```
gtanzillo, line 144: What should happen if there was an attempt to change `type` to something else in an edit request? Should we test for that case?

gtanzillo (approved):
Looks good 👍

qpc/cred/test_openshift_cred_edit.py
nicolearagao: An error will be raised. I can add the test for this scenario.

qpc/cred/test_openshift_cred_edit.py
```diff
@@ -162,3 +162,35 @@ def test_edit_incorrect_arg(
         out, err = capsys.readouterr()
         assert out == ""
         assert "unrecognized argument" in err
+
```
nicolearagao, line 165: @gtanzillo

gtanzillo, line 165: ❤️

PR #206, opened 2022-10-19
Feature/openshift add source

nicolearagao:
Add openshift source to CLI.

This pr is cli implementation of task [DISCOVERY-186].

### DIFF
--- a/qpc/source/conftest.py
+++ b/qpc/source/conftest.py
@@ -1,0 +1,31 @@
+# Copyright (c) 2022 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+#
+#  pylint: disable=import-outside-toplevel
+"""pytest configuration file."""
+
+import pytest
+
+from qpc.cred import CREDENTIAL_URI
+
+
+@pytest.fixture(autouse=True)
+def _setup_server_config_file(server_config):
+    ...
+
+
+@pytest.fixture
+def ocp_credential_mock(requests_mock):
+    """Return ocp credential."""
+    from qpc.utils import get_server_location
+
+    get_cred_url = get_server_location() + CREDENTIAL_URI + "?name=ocp_cred_1"
+    cred_results = [{"id": 1, "name": "ocp_cred_1"}]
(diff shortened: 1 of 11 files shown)


qpc/source/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+#  pylint: disable=import-outside-toplevel
+"""pytest configuration file."""
+
+import pytest
+
+from qpc.cred import CREDENTIAL_URI
+
+
+@pytest.fixture(autouse=True)
+def _setup_server_config_file():
```
gtanzillo, line 19: Since you have the same server setup here and in the credential test, I'm thinking you could DRY this up a bit if you make `conftest.py` global by moving it up a directory so that it can be included in those 2 tests and perhaps others. Just a suggestion

nicolearagao: I have thought about it. Actually, this is used in other places of the code as well. The reason why I didn't move it there yet is because it would involve a little refactor on other tests so I was leaving it for the next quarter. However there's no re4ason to not do it know and and refactor 
 it later.

bruno-fs: I like the idea of DRYing it up and we can do that without refactoring other tests
(and I wouldn't advise having this particular one set to be autoused globally even if
we had all tests refactored.)


```python
# toplevel conftest.py

@pytest.fixture
def server_config():
    from qpc.utils import write_server_config

    return write_server_config(
        {
            "host": "127.0.0.1",
            "port": 8000,
            "use_http": True,
            "require_token": False,
        }
    )

# qpc/*/conftest.py

@pytest.fixture(autouse=True)
def _server_config(server_config):
    ...

```

This will dry it up and also give us flexibility to enable it whenever we see fit.

bruno-fs (commented):
LGTM! :+1:
But I would dry up the server config fixture (see the suggestion, I reopened the discussion).

qpc/source/utils.py
```diff
@@ -70,13 +70,11 @@ def build_source_payload(args, add_none=True):
...
     # Add source options
     options = None
-    if (hasattr(args, 'ssl_cert_verify') and
-            args.ssl_cert_verify is not None):
-        ssl_cert_verify = args.ssl_cert_verify.lower() == 'true'
-        options = {'ssl_cert_verify': ssl_cert_verify}
-    if (hasattr(args, 'disable_ssl') and
-            args.disable_ssl is not None):
-        disable_ssl = args.disable_ssl.lower() == 'true'
+    if hasattr(args, "ssl_cert_verify") and args.ssl_cert_verify is not None:
```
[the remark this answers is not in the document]
bruno-fs: since changing this would make it stylistically distinct from the other statements, let's leave as is.

bruno-fs (approved):
LGTM :+1:

gtanzillo (approved):

PR #207, opened 2022-10-31
Feature/password prompt insights add_login

nicolearagao:
- Test improvements on qpc insights test modules
- The user will be prompted for his password instead of pasting it directly in terminal.

This implements [DISCOVERY-200]

### DIFF
--- a/qpc/insights/conftest.py
+++ b/qpc/insights/conftest.py
@@ -18,3 +18,10 @@ def _setup_server_config_file(server_config):
     # We wanted the fixture to be widely available,
     # but only called when necessary as we use autouse flag
 
+
+@pytest.fixture
+def insights_login_password_input(monkeypatch):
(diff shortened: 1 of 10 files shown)


qpc/insights/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+#
+"""pytest configuration file."""
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _setup_server_config_file(server_config):
```
bruno-fs, line 16: I think its a good idea to add a brief comment explaining
1) this is disabling another fixture
2) why we are doing this

bruno-fs (commented):
LGTM

gtanzillo (approved):
LGTM 👍

PR #209, opened 2022-11-03
Fix/type in cred list should be case insensitive

nicolearagao:
Accept case insensitive args in cred list --type


gtanzillo (approved):
Good change! 👍

bruno-fs (approved):
LGTM :+1:

ruda (approved):
OK!

PR #215, opened 2023-02-01
fix: drop unnecessary tox plugin

bruno-fs:
tox-poetry is no longer required. As of tox>=4, tox and poetry play nice with each other.
See https://python-poetry.org/docs/faq/#is-tox-supported

Also drop support for python 3.7 (it's close to reach EOL and would require extra work to mantain dependencies) and add new python versions.


nicolearagao (approved):
LGTM 👍🏼

PR #221, opened 2023-03-14
CLI shouldn't force output file for json reports 

nicolearagao:
Output files won't be required for deployments and details 
reports, content will be written to stdout as default when the user
choose json file format.
For insights reports, the behavior will be the same except for a little detail, the --json flag won't be necessary, the json report will be directly written to the stdout. 

This implements [[DISCOVERY-184]](https://issues.redhat.com/browse/DISCOVERY-184)

PR #222, opened 2023-03-23
Refactor/collect logs appropriately

nicolearagao:
waiting on https://github.com/quipucords/qpc/pull/221


qpc/cred/add.py
```diff
@@ -22,6 +11,8 @@
 from qpc.request import POST
 from qpc.translation import _
 
+log = getLogger("qpc")
```
infinitewarp, line 14: I noticed that all of your loggers are "qpc". Is there a reason specifically for them all to be the same? If we use `__name__` instead (see suggested change), then in the future if we want finer control, we could adjust each package or module's log level _independently_ in the project-wide logging config. If they're all using the same logger name like "qpc", though, you have to change all or none.

Also, though this is more of a style choice, I think the prevailing convention is to call the Logger object `logger`. Of course, if you do that, be sure to replace any `log` references with `logger`.

```suggestion
logger = getLogger(__name__)
```

See also: https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial

> A good convention to use when naming loggers is to use a module-level logger, in each module which uses logging, named as follows:
> 
> `logger = logging.getLogger(__name__)`
> 
> This means that logger names track the package/module hierarchy, and it’s intuitively obvious where events are logged just from the logger name.

qpc/cred/add.py
```diff
@@ -128,4 +119,4 @@ def _build_data(self):
         self.req_payload = build_credential_payload(self.args, self.args.type)
 
     def _handle_response_success(self):
-        print(_(messages.CRED_ADDED % self.args.name))
+        log.info(_(messages.CRED_ADDED % self.args.name))
```
infinitewarp, line 122: A couple "best practices" suggestions here! 😅 

First, if we're using gettext translations (the `_` function), then we should never pass variables like `self.args.name` into it because that would result in a whole new string to translate _including_ the value of that variable, for every variable it might encounter. Instead, we should always translate only "tokenized" strings with variable placeholders. You then format the variables into the string _after_ translation has occurred (after the `_` function executes).

For example, if I want `"My name is %s" % name` to be translated correctly, I would use `_("My name is %s") % name`.

If you have a string that takes _multiple_ variables, it's also important also to _name_ the tokens. Not all language grammars use the same word order, and using tokens enables the translator to rearrange the content correctly for that language. Think about how some languages typically use adjective-noun but other languages typically use noun-adjective.

For example, `_("The %(color)s %(food)s is %(quality)s.") % {"color": "red", "food": "apple", "quality": "delicious"}`.

Second, instead of using `%` to format the string, we should pass the tokenized string and its variables as separate arguments to the `logger.info` call. Why? Well, if you use `%` that means the string is always _immediately_ formatted _before_ sending it to the logger's function. However, you might have a logging configuration that isn't going to output a particular call. Consider, for example, that we often don't _actually_ output `logger.debug` calls. If you simply pass the arguments to the logger function, it will only format the string _if necessary_. If the logging config would skip that call (like `debug`), then it also skips formatting, which saves memory and CPU time. It doesn't save a _lot_, especially for a small message like this, but it adds up, and it's a good habit to follow in general.

Extending the previous example, that would look like:

```python
logger.info(
    _("The %(color)s %(food)s is %(quality)s."),
    {"color": "red", "food": "apple", "quality": "delicious"}
)
```

(Whitespace added to make the separation obvious.)

```suggestion
        logger.info(_(messages.CRED_ADDED), self.args.name)
```

qpc/cred/clear.py
```diff
@@ -24,6 +12,8 @@
 from qpc.translation import _
 from qpc.utils import handle_error_response
 
+log = getLogger("qpc")
```
infinitewarp, line 15:
```suggestion
logger = getLogger(__name__)
```

qpc/cred/clear.py
```diff
@@ -71,26 +61,26 @@ def _delete_entry(self, credential_entry, print_out=True):
         if response.status_code == codes.no_content:
             deleted = True
             if print_out:
-                print(_(messages.CRED_REMOVED % name))
+                log.info(_(messages.CRED_REMOVED % name))
```
infinitewarp, line 64:
```suggestion
                logger.info(_(messages.CRED_REMOVED), name)
```

qpc/cred/clear.py
```diff
@@ -71,26 +61,26 @@ def _delete_entry(self, credential_entry, print_out=True):
...
         else:
             handle_error_response(response)
             if print_out:
-                print(_(messages.CRED_FAILED_TO_REMOVE % name))
+                log.error(_(messages.CRED_FAILED_TO_REMOVE % name))
```
infinitewarp, line 68:
```suggestion
                logger.error(_(messages.CRED_FAILED_TO_REMOVE), name)
```

qpc/messages.py
```diff
@@ -300,6 +300,7 @@
 
 MERGE_JOB_ID_NOT_FOUND = "Report merge job %s not found."
 MERGE_JOB_ID_STATUS = "Report merge job %s is %s."
+MERGE_ERROR = "No reports found.  Error json: %s"
```
bruno-fs, line 303: Do we truly want 2 spaces here?

bruno-fs (commented):
The way I see, this PR has 2 very distinct intentions:
- remove copyright headers
- replace print statements with logs

In the future, please consider opening separate PRs in situations like this

qpc/utils.py
```diff
@@ -465,7 +463,7 @@ def read_in_file(filename):
                 result = in_file.read().splitlines()
         except EnvironmentError as err:
             err_msg = t(messages.READ_FILE_ERROR % (input_path, err))
-            log.error(err_msg)
+            logger.error(err_msg)
```
infinitewarp, line 466: It looks like `t` is just a renamed `_` (the standard translation function), and it looks like `err_msg` is only used once here. So, to follow the new pattern we've established throughout this PR, I think these lines should also change to something like:

```python
logger.error(t(messages.READ_FILE_ERROR), input_path, err)
```

And if you have time and aren't exhausted from these changes yet, you could also update the definition of `READ_FILE_ERROR` to include placeholder tokens like this:

```python
READ_FILE_ERROR = "Error reading from %(path)s: %(error)s."
```

…which could result in an even better:

```python
logger.error(t(messages.READ_FILE_ERROR), {"path": input_path, "error": err})
```

qpc/scan/clear.py
```diff
@@ -70,11 +71,11 @@ def _delete_entry(self, scan_entry, print_out=True):
         if response.status_code == codes.no_content:
             deleted = True
             if print_out:
-                print(_(messages.SCAN_REMOVED % name))
+                logger.info(_(messages.SCAN_REMOVED), name)
         else:
             handle_error_response(response)
             if print_out:
-                print(_(messages.SCAN_FAILED_TO_REMOVE % name))
+                logger.error(_(messages.SCAN_FAILED_TO_REMOVE % name))
```
infinitewarp, line 78: Missed this one!
```suggestion
                logger.error(_(messages.SCAN_FAILED_TO_REMOVE), name)
```

infinitewarp (changes requested):
The logger commands looks _soooo_ much nicer now. Thank you @nicolearagao for giving them so much attention. I spotted just a couple that you may have missed, but that's nothing compared to the _million_ that you did perfectly. Once those are fixed, I think this puppy is ready to merge. 😀

qpc/utils.py
nicolearagao: I tried to get rid of this t in the past for consistency in code but it broke a lot of tests. I will address this in a later opportunity.

infinitewarp (approved):
Great work! 👏 🚀

PR #227, opened 2023-04-16
fix: stop storing binary responses in qpc.log

nicolearagao:
Any binaries received in response_json that can be converted to json will be stored, otherwise replaced by "encoded blob ignored"

### DIFF
--- a/qpc/request.py
+++ b/qpc/request.py
@@ -143,6 +143,15 @@ def put(url, payload, headers=None):
     return requests.put(url, json=payload, headers=headers, verify=ssl_verify)
 
 
+methods = {
+    "POST": post,
+    "GET": get,
+    "PATCH": patch,
+    "DELETE": delete,
+    "PUT": put,
+}
+
+
 # pylint: disable=too-many-arguments, too-many-branches, too-many-locals
 def request(
     method,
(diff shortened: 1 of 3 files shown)


qpc/utils.py
```diff
@@ -384,6 +384,8 @@ def log_request_info(method, command, url, response_json, response_code):
     :param response_json: the response returned from the request
     :param response_code: the status code being returned (ie. 200)
     """
+    if isinstance(response_json, (bytes, bytearray)):
+        response_json = response_json.decode("utf-8", errors="ignore")
```
bruno-fs, line 388: some responses won't make sense. even decoded.

Have you tested this with insights tarball and/or "all reports" tarball?

bruno-fs (changes requested):
The presented solution won't solve the original problem.

qpc/utils.py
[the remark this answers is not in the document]
bruno-fs: The original issue appeared with not just compressed data: it appeared with insights report, which consists of a gzipped tarball containing at least 2 files.

bruno-fs:
can we close this one @nicolearagao ?

nicolearagao:
> can we close this one @nicolearagao ?

I will push changes

qpc/request.py
```diff
@@ -169,58 +169,86 @@ def request(
...
-        elif method == PUT:
-            result = handle_general_errors(
-                put(url, payload, req_headers), min_server_version
+        methods = {
+            "POST": post,
+            "GET": get,
+            "PATCH": patch,
+            "DELETE": delete,
+            "PUT": put
+        }
```
infinitewarp, line 185: Minor suggestion: since `methods` is always being set to the same value every time `request` is called, and nothing updates `methods`, we could move its assignment to before the `request` function (to lines 144-145).

This would also shrink the `try` block, which should only contain code that we need to handle exceptions. Even if we don't move the `methods` assignment outside of the function, we should at least move it to before the `try` (to line 177) since this assignment should never raise a `ConnectionError` or `SSLError` exception.

qpc/request.py
```diff
@@ -169,58 +169,86 @@ def request(
...
+                method,
+                request_method,
+                url,
+                params,
+                payload,
+                req_headers,
+                min_server_version
             )
-        else:
+        except KeyError:
```
infinitewarp, line 197: Another suggestion, though this one falls into the highly subjective debate of [EAFP](https://docs.python.org/3/glossary.html?highlight=eafp#term-EAFP) versus [LBYL](https://docs.python.org/3/glossary.html?highlight=eafp#term-LBYL).

I think I would prefer LBYL in this case because it flattens the structure and makes it easier to read. If we explicitly check for the presence of `method` first, we could remove one layer of these double-nested `try` blocks. That might look like:

```python
if method not in methods:
    logger.error("Unsupported request method %s", method)
    parser.print_help()
    sys.exit(1)

try:
    request_method = methods[method]
    result = perform_request(...)
except (requests.exceptions.ConnectionError, requests.exceptions.SSLError):
    handle_connection_error() 

log_request_info(...)
return result
```

Note that I also moved `log_request_info` and `return` outside of the `try` block since we aren't handling any exceptions that they would raise (I think).

infinitewarp: Alternatively, if you prefer EAFP, we could totally separate the `try`/`except` blocks like:

```python
try:
    request_method = methods[method]
except KeyError:
    logger.error("Unsupported request method %s", method)
    parser.print_help()
    sys.exit(1)

try:
    result = perform_request(...)
except (requests.exceptions.ConnectionError, requests.exceptions.SSLError):
    handle_connection_error() 

log_request_info(...)
return result
```

qpc/request.py
```diff
@@ -169,58 +169,86 @@ def request(
...
+        logger.error(
+            _(CONNECTION_ERROR_MSG),
+            {"protocol": protocol, "host": host, "port": port}
+        )
+    logger.error(_(messages.SERVER_CONFIG_REQUIRED), PKG_NAME)
+
+
+def perform_request(
+    method,
+    request_method,
```
infinitewarp, line 228: Hmm…

It feels weird that `request` is using `method` to get `request_method`, and then it just passes _both_ of those into `perform_request`, and `request` does nothing else with the `request_method` that it just created. This makes me think there's no real benefit to instantiating `request_method` outside of `perform_request`, and `request` should send just `method` to `perform_request`, and `perform_request` should do the `request_method` lookup.

Beyond the scope of this change (and I am _not_ suggesting you do the following in _this_ PR), I would like to strip away some unnecessary layers of abstraction in this module. Our current `get`, `post`, etc. functions are doing nothing useful. They just add complication to functions like this. 😕

infinitewarp: Sorry if what I said there seems incoherent. I'm not sure if I clearly expressed my confusing line of thoughts. 🤔 

I want to revisit all this code _in general_ some time later, not just what you've written in this PR, to straighten out the seemingly useless `get`, `post`, etc. functions, and I expect most of `request_method` would disappear if/when that happens. So, my suggestion to move the `request_method` lookup from `request` to `request_method` would technically be an improvement, but it's a _very minor_ one, and not something that should block the PR in its current state.

nicolearagao: OK, I understand. I will work on a better solution applying your suggestions and push a fixup.

infinitewarp (approved):

PR #233, opened 2023-05-05
feat: adding support for building and running the qpc cli from an image

abellotti:
feat: adding support for building and running the qpc cli from an image

- Adding the Dockerfile
- Adding the deploy/ docker_run.sh entrypoint
- Adding the make build-container directive

rpm/toolbox is very rhel centric, was having my M1 mac feeling left out.  This will allow us to run qpc from anywhere and we can include the man pages in this image which accompanies the CLI.

Usage:

```
$ make build-container
$ alias qpc="podman run -it -v ~/.config/qpc:/root/.config/qpc \
                            -v ~/.local/share/qpc:/root/.local/share/qpc quipucords-cli"
$ qpc server config --host discovery-server-discovery-qe.apps.discocluster.lab.upshift.rdu2.redhat.com --port 443
$ qpc server login --username admin
Password:
$ qpc source list
[
  {
     "connection": {
          ...
  }
]
```

Accessing the man page:

```
$ qpc man qpc
qpc(1)                                              QPC Command Line Guide                                             qpc(1)

Name
       qpc  - Inspect and report on product entitlement metadata from various sources, including networks and systems manage-
       ment solutions.

Synopsis
       qpc command subcommand [options]

Description
       The Quipucords tool, accessed through the qpc command, is an inspection and reporting tool.  It is designed to identi-
       fy  environment  data, or facts, such as the number of physical and virtual systems on a network, their operating sys-
       tems, and other configuration data.  In addition, it is designed to identify and report more detailed facts  for  some
       versions  of  key  Red Hat packages and products for the Linux based IT resources in that network.  The ability to in-
       spect the software and systems that are running on your network improves your ability to understand and report on your
       entitlement usage.  Ultimately, this inspection and reporting process is part of the larger system administration task
       of managing your inventories.

       The Quipucords tool uses two types of configuration to manage the inspection process.  A credential contains  configu-
       ration  such  as the user name and password or SSH key of the user that runs the inspection process.  A source defines
       the entity to be inspected and one or more credentials to use during the inspection process.  The  entity  to  be  in-
       spected  can  be a host, subnet, network, or systems management solution such as vCenter Server or Satellite.  You can
       save multiple credentials and sources to use with Quipucords in various combinations as you run inspection  processes,
       or scans.  When you have completed a scan, you can access the output as a report to review the results.

       By  default, the credentials and sources that are created when using Quipucords are encrypted in a database.  The val-
       ues are encrypted with AES-256 encryption.  They are decrypted when the Quipucords server runs a scan by using a vault
       password to access the encrypted values that are stored in the database.

       The  Quipucords tool is an agentless inspection tool, so there is no need to install the tool on the sources to be in-
       spected.

       This manual page describes the commands, subcommands, and options for the qpc command and includes  usage  information
       and example commands.

Usage
       The  qpc  command has several subcommands that encompass the inspection and reporting workflow.  Within that workflow,
 Manual page qpc(1) line 1 (press h for help or q to quit)
```


gtanzillo:
@abellotti I just tried this out on my local machine and it works like a champ!!

deploy/docker_run.sh
```diff
@@ -0,0 +1,22 @@
...
+shift
+if [ "${QPC_COMMAND}" == "qpc" ]; then
+  poetry run qpc "${@}"
+elif [ "${QPC_COMMAND}" == "man" ]; then
+  man "${@}"
+else
+  echo "Unknown command ${QPC_COMMAND} specified."
+  echo ""
+  echo "Usage:"
+  echo "   $ alias dsc=\"docker run -it -v ~/.config/qpc:/root/.config/qpc \\"
```
bruno-fs, line 13:
```suggestion
  echo "   $ alias qpc=\"docker run -it -v ~/.config/qpc:/root/.config/qpc \\"
```
since this is upstream, we should refer to cli as `qpc`.

deploy/docker_run.sh
```diff
@@ -0,0 +1,22 @@
...
+  echo "                               -v ~/.local/share/qpc:/root/.local/share/qpc \\"
+  echo "                               quipucords-cli\""
+  echo "   $ dsc <command> <args>"
```
bruno-fs, line 16:
```suggestion
  echo "   $ qpc <command> <args>"
```

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+        python3.11 \
+        python3.11-pip \
+        tar \
+        which \
+    && if [[ ! -e /usr/bin/python ]]; then ln -sf /usr/bin/python3.11 /usr/bin/python; fi \
+    && if [[ ! -e /usr/bin/pip    ]]; then ln -sf /usr/bin/pip3.11    /usr/bin/pip;    fi \
+    && pip install -U pip \
+    && pip install poetry \
+    && poetry config virtualenvs.in-project true \
+    && poetry install -n
```
bruno-fs, line 26: poetry install will install all dev dependencies as well.
I'd install only the main ones
```suggestion
    && poetry install -n --only main
```

deploy/dsc
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
bruno-fs, line 1: since this is upstream, the script should be named "qpc"

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+ENV VIRTUAL_ENV=/app/qpc/.venv
+ENV PATH="${VIRTUAL_ENV}/bin:$PATH"
+
+# Copy QPC
+COPY . .
+COPY docs/qpc.1 /usr/local/share/man/man1/qpc.1
```
bruno-fs, line 33: Aren't we missing a `make manpage` before copying the file here?

bruno-fs (commented):
Nice job! Few comments to improve - most important ones are on the dockerfile

This version doesn't work on toolbox (it requires some extra dependencies and using CMD instead of ENTRYPOINT), but I'm not sure if we still want it - if we do want it, toolbox support can be added later.

Dockerfile
abellotti: It's in the Makefile dependency for the build-container target.

```
build-container: manpage
  podman build ....
```

bruno-fs: aaah I did miss that part! cool

bruno-fs (approved):

PR #234, opened 2023-05-05
fix: improve acceptable arg options text

nicolearagao:
The fields in which this text is being used were not case insensitive in the past, therefore needed a more explicit description on accepted values. A recent refactor modified this behavior.

This implements [[DISCOVERY-206]](https://issues.redhat.com/browse/DISCOVERY-206)

### DIFF
--- a/qpc/source/__init__.py
+++ b/qpc/source/__init__.py
@@ -22,5 +22,5 @@ SOURCE_TYPE_CHOICES = [
     VCENTER_SOURCE_TYPE,
 ]
 
-BOOLEAN_CHOICES = ["True", "False", "true", "false"]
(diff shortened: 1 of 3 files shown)


infinitewarp (approved):
In isolation, this is a fine improvement. I approve!

However, for the future of `qpc`, I think we should drop the boolean strings in favor of `argparse`'s built-in bool handling. Calling `add_argument` with `action="store_true"` is the generally accepted best practice here, and `action=argparse.BooleanOptionalAction` is also available if for separate explicit true/false settings. Either of these would be a general improvement, but they would probably require updating camayoc, and I would not suggest making that change with _this_ PR.

PR #239, opened 2023-05-24
test: adjust tests for python 3.11

bruno-fs:
on python 3.11 argparser can't have subcommands overriden, something that was happening on the refactored tests.


nicolearagao (approved):
LGTM

PR #245, opened 2023-06-05
Feature/add cmd for rebranding qpc into dsc

nicolearagao:
- Add jinja-cli dependency to build poetry group
- read environment variable for package source info
- create jinja template for manpage
- add manpage generation to manpage command

This feature implements [[DISCOVERY-310]](https://issues.redhat.com/browse/DISCOVERY-310)


abellotti:
Nice to see Jinja making its appearance 👍  Is the .rst file still needed in the repo since we're regenerating it from the .j2 file before pandoc is invoked  ?

nicolearagao:
> Nice to see Jinja making its appearance +1 Is the .rst file still needed in the repo since we're regenerating it from the .j2 file before pandoc is invoked ?

It won't be necessary. I wanted to take a second look at the report generated to make sure everything looked ok. I will remove the file.

docs/source/man.rst
```diff
@@ -808,7 +808,8 @@ The authentication data in the credentials and the network-specific and system-s
 Authors
 -------
 
-Quipucords is written and maintained by Red Hat. Please refer to the commit history for a full list of contributors.```
```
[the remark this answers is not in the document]
bruno-fs: 100%. Either rst goes away, or j2 goes away.

I'd vote for a rename, tho (j2 -> rst as a jinja template)

bruno-fs (changes requested):
There's absolutely no need for 2 files.

qpc/release.py
```diff
@@ -1,9 +1,11 @@
...
+
 from . import __package__version__
 
 VERSION = __package__version__
 AUTHOR = "QPC Team"
 AUTHOR_EMAIL = "qpc@redhat.com"
-PKG_NAME = "qpc"
-ENTRYPOINT = f"{PKG_NAME}=qpc.__main__:main"
+QPC_VAR_PKG_SOURCE = os.environ.get("QPC_VAR_PKG_SOURCE", "qpc")
+ENTRYPOINT = f"{QPC_VAR_PKG_SOURCE}=qpc.__main__:main"
```
bruno-fs, line 10: the commit that introduce this refactor is a "refactor" not "feat".

docs/source/man.j2
```diff
@@ -0,0 +1,816 @@
+{{ QPC_VAR_PKG_SOURCE|default("qpc", true) }}
```
bruno-fs, line 1: This is valid for all variables:

I don't see why to set the default value each time this is called.
Why not simply set the default in the script (or make role) that will generate the man.rst?

The readability will be slightly improved and the maintenance will be GREATLY improved.

bruno-fs: (this is valid for the whole file and all variables)

Makefile
```diff
@@ -51,13 +54,26 @@ test-coverage:
 	poetry run coverage report --show-missing
 	poetry run coverage xml
 
+generate-man:
+	@export QPC_VAR_CURRENT_YEAR=$(CURRENT_YEAR) \
```
bruno-fs, line 58: This works, why `CURRENT_YEAR` is not prefixed with `QPC_VAR_`? That would be simpler and cleaner...

Makefile
```diff
@@ -51,13 +54,26 @@ test-coverage:
...
+	&& jinja -X QPC_VAR docs/source/man.j2 $(ARGS)
+
+update-man.rst:
+	$(MAKE) generate-man ARGS="-o docs/source/man.rst"
+
+manpage-test:
+	@poetry run $(MAKE) generate-man | grep -v "Entering directory" | \
+	grep -v "Leaving directory" > output.txt
+	@diff -u docs/source/man.rst output.txt
```
bruno-fs, line 67: why manpage-test neesds these "grep -v"s and manpage role doesn't?

nicolearagao: good observation. It also needs to strip those undesired lines. I made the correction.

docs/source/man.rst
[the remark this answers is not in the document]
nicolearagao: @brad that was clear, thanks anyway

Makefile
```diff
@@ -51,13 +53,29 @@ test-coverage:
...
+	@export QPC_VAR_CURRENT_YEAR=$(shell date +'%Y') \
+	&& export QPC_VAR_PROJECT=$${QPC_VAR_PROJECT:-Quipucords} \
+	&& export QPC_VAR_PKG_SOURCE=$${QPC_VAR_PKG_SOURCE:-qpc} \
+	&& jinja -X QPC_VAR docs/source/man.j2 $(ARGS)
+
+update-man.rst:
+	$(MAKE) generate-man ARGS="-o docs/source/man.rst"
+
+manpage-test:
+	@poetry run $(MAKE) generate-man | grep -v "Entering directory" | \
```
infinitewarp, line 66: I have an idea that may negate the need for these negative greps. Putting this comment here as a placeholder until I have tested my idea so we don't forget.

infinitewarp: Actually, now that I try it locally, I don't think _any_ of the `grep`s are necessary. 🤔

I changed this locally:

```diff
diff --git a/Makefile b/Makefile
index b5ca073..6ee028b 100644
--- a/Makefile
+++ b/Makefile
@@ -63,13 +63,11 @@ update-man.rst:
        $(MAKE) generate-man ARGS="-o docs/source/man.rst"

 manpage-test:
-       @poetry run $(MAKE) generate-man | grep -v "Entering directory" | \
-       grep -v "Leaving directory" > output.txt
-       @diff -u docs/source/man.rst output.txt
+       @poetry run $(MAKE) generate-man | diff -u docs/source/man.rst -

 manpage:
-       @$(MAKE) generate-man | grep -v "Entering directory" | \
-       grep -v "Leaving directory" | \
+       @$(MAKE) generate-man | \
        $(pandoc) -s - \
          --standalone -t man -o docs/$(QPC_VAR_PKG_SOURCE).1\
          --variable=section:1\
```

But I do not see "Entering directory" or "Leaving directory" anywhere in the output or generated files. Maybe it used to appear but no longer appears due to some other change? I recommend removing the `grep`s like this and double-check that the commands still complete as expected.

(Note that I also removed the intermediary `output.txt` file so that running these commands will not leave a dirty/untracked file in the local repository.)

Originally, when I assumed that "Entering directory" or "Leaving directory" was still being generated somewhere, I was going to suggest changing how we call `make` from within `make`, because `make` _itself_ would be responsible for those messages. The `--no-print-directory` argument would tell `make` to silence these specific informational messages. Alternatively, the `-s` argument would tell `make` to silence _all_ make-related messages.

I wonder if you and I have something different in our environments that is causing `make` to emit the directory changes for you but not for me. I wonder if you have any `MAKEFLAGS` set in your shell environment.

Makefile
```diff
@@ -51,13 +53,29 @@ test-coverage:
...
-	  --variable=footer:'version $(PKG_VERSION)' \
-	  --variable=header:'QPC Command Line Guide'
+	@$(MAKE) generate-man | grep -v "Entering directory" | \
+	grep -v "Leaving directory" | \
+	$(pandoc) -s - \
+	  --standalone -t man -o docs/$(QPC_VAR_PKG_SOURCE).1\
+	  --variable=section:1\
+	  --variable=date:'$(BUILD_DATE)'\
+	  --variable=footer:'version $(PKG_VERSION)'\
+	  --variable=header:'$(QPC_VAR_PKG_SOURCE_UPPER) Command Line Guide'
```
infinitewarp, line 78: Should the name really be forced to upper-case in strings like this?

Asking because I do not know the brand standards here, but changing the case of a program name based on documentation context feels strange to me. Maybe this is a question for @bluemorpho-mpw?

Makefile
```diff
@@ -2,6 +2,8 @@ DATE		= $(shell date)
 PYTHON		= $(shell which python)
 PKG_VERSION = $(shell poetry -s version)
 BUILD_DATE  = $(shell date +'%B %d, %Y')
+QPC_VAR_PKG_SOURCE := $(or $(QPC_VAR_PKG_SOURCE), "qpc")
```
infinitewarp, line 5: Where did the name `QPC_VAR_PKG_SOURCE` come from?

At a glance, I wouldn't know what `QPC_VAR_PKG_SOURCE` means if I'm not already familiar with this code. I understand `QPC_` is a prefix we want to standardize, but `VAR_PKG_SOURCE` doesn't have an intrinsic meaning to me. It may be easy to understand if you find the original definition _here_, but I wonder if a name with a more direct meaning would make it more readable in the docs and code. Maybe a name like `QPC_PROGRAM_NAME` is more precise assuming it is always meant to refer to the program `qpc` or `dsc`.

Thoughts?

infinitewarp (changes requested):
I posted a few questions inline, but otherwise this looks good to me! Once we answer those, let's get this merged. 🙂

Makefile
bruno-fs: Mmm worth a shot. I did reproduce a issue she was having locally and suggested writing to a temporary file to avoid these negative greps. I prefer your solution, though 👍🏽

Makefile
infinitewarp: See also a needlessly pedantic discussion in Slack here: https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1687550700970519

I'm sorry I asked! 😂 

Feel free to resolve this conversation and keep what we have if no strong recommendations come from that thread.

Makefile
nicolearagao: PKG_SOURCE was the original name, and I didnt think about changing it, Ive just added a prefix to be able to use a regex when importing env variables. QPC_VAR_PROGRAM_NAME sounds good, I will make the changes.

Makefile
nicolearagao: I dont think we reached a conclusion there, I will resolve this.

Makefile
nicolearagao: I will push the changes.

nicolearagao: ok, I will try to make it work locally based on your suggestion.

nicolearagao: Just confirming that "Entering directory" or "Leaving directory" messages are still being generated for me locally, I am not sure why, I have no `MAKEFLAGS` set on my shell. Since this behavior was also reproduced locally in @bruno-fs 's env I believe the best approach would be calling the `make `command with the `--no-print-directory`, that works well. You mentioned that -s would be an alternative, but wouldn't this flag also suppress any error messages related to the command generate-man ? In the end we would still reach our goal and have the test fail, but wouldn't it be interesting to see why the command failed in the first place ?

infinitewarp: You are correct. `--no-print-directory` is probably the better choice here because, yes, `-s` may more aggressively suppress other messages we do want to see.

infinitewarp (approved):
Looks good! 🎉

Makefile
```diff
@@ -63,7 +63,7 @@ update-man.rst:
 	$(MAKE) generate-man ARGS="-o docs/source/man.rst"
 
 manpage-test:
-	@poetry run $(MAKE) --no-print-directory generate-man | diff -u docs/source/man.rst -
```
infinitewarp: Don't forget to put this back, assuming this was a temporary removal given the commit was called "test".

qpc/release.py
```diff
@@ -1,9 +1,11 @@
...
+import os
+
 from . import __package__version__
 
 VERSION = __package__version__
 AUTHOR = "QPC Team"
 AUTHOR_EMAIL = "qpc@redhat.com"
-PKG_NAME = "qpc"
-ENTRYPOINT = f"{PKG_NAME}=qpc.__main__:main"
+QPC_VAR_PKG_SOURCE = os.environ.get("QPC_VAR_PKG_SOURCE", "qpc")
```
bruno-fs, line 9: why not `QPC_VAR_PKG_NAME`

bruno-fs: or, even better, `QPC_VAR_PROGRAM_NAME` as it is declared on the jinja template?

docs/source/man.j2
```diff
@@ -0,0 +1,816 @@
...
+
+Synopsis
+--------
+
+``{{ QPC_VAR_PROGRAM_NAME }} command subcommand [options]``
+
+Description
+-----------
+
+The {{ QPC_VAR_PROJECT }} tool, accessed through the ``{{ QPC_VAR_PROGRAM_NAME }}`` command, is an inspection and reporting tool. It is designed to identify environment data, or *facts*, such as the number of physical and virtual systems on a network, their operating systems, and other configuration data. In addition, it is designed to identify and report more detailed facts for some versions of key Red Hat packages and products for the Linux based IT resources in that network. The ability to inspect the software and systems that are running on your network improves your ability to understand and report on your usage. Ultimately, this inspection and reporting process is part of the larger system administration task of managing your inventories.
```
bruno-fs, line 18: I understand quipucords/discovery are not actual products, but for the purpose of this document it could be clearer if we use this nomenclature (`QPC_VAR_PRODUCT`). Thoughts @infinitewarp @nicolearagao ?

bruno-fs: or maybe `QPC_VAR_SERVER_NAME` to be more explicit 🤔

infinitewarp: `QPC_VAR_PROJECT` resolves to `Quipucords`, not `qpc`.

I believe the current implementation by @nicolearagao matches the behavior of the existing implementation. Do we really want to change that here? I think that should be a separate discussion and PR if so. Changing that value would fall under @bluemorpho-mpw's expertise.

qpc/release.py
nicolearagao: agreed!

docs/source/man.j2
nicolearagao: ++ to discussion it further later on and modifying it with a supplemental pr.

bruno-fs (approved):
LGTM

PR #246, opened 2023-06-06
chore: replacing isort with the isort rule enablement in ruff

abellotti:
chore: replacing isort with the isort rule enablement in ruff

- Removing the lint-isort directive
- Adding the "I" isort rule in ruff
- Fixing one inconsistency between the two
- Removing the isort package dependency from poetry


setup.py
```diff
@@ -4,9 +4,8 @@
 import sys
 from pathlib import Path
 
-from setuptools import find_packages, setup
-
 from qpc.release import AUTHOR, AUTHOR_EMAIL, ENTRYPOINT, PKG_NAME, URL, VERSION
+from setuptools import find_packages, setup
```
bruno-fs, line 8: huh, this is weird. qpc should be in the last section of imports 🤔

bruno-fs: ah, we need a ruff section for isort, configuring qpc as known first party!

bruno-fs: https://beta.ruff.rs/docs/settings/#known-first-party

abellotti: Ahh, thank you.  I'll give that a try.

infinitewarp (approved):

bruno-fs (approved):

PR #247, opened 2023-06-15
feat: Add support for the ssh_keyvalue parameter for credentials

abellotti:
feat: Add support for the ssh_keyvalue parameter for credentials

- Allows us to specify an ssh_keyvalue instead of a password or ssh_keyfile.
- It is supported for both credential creation and edits.

This capability enhances support for Quipucords https://github.com/quipucords/quipucords/pull/2390


bruno-fs (commented):
I like the approach, I couldn't think of a better way to handle the limitation of running the CLI on a container. 

Unfortunately the UX is not great, but honestly I think this is the best we can have while distributing the CLI as a container...

qpc/cred/utils.py
```diff
@@ -72,3 +134,22 @@ def build_credential_payload(args, cred_type, add_none=True):
...
+    sys.stderr.flush()
+    stdin_fd = sys.stdin.fileno()
+    orig_tios = termios.tcgetattr(stdin_fd)
+    noecho_tios = termios.tcgetattr(stdin_fd)
+    noecho_tios[3] = noecho_tios[3] & ~termios.ECHO
+    try:
+        termios.tcsetattr(stdin_fd, termios.TCSADRAIN, noecho_tios)
+        multiline_password = sys.stdin.readlines()
+    finally:
+        termios.tcsetattr(stdin_fd, termios.TCSADRAIN, orig_tios)
```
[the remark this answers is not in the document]
abellotti: No problem being verbose, I'll update 👍 . I didn't want to diverge too much from the Python provided example here:  https://docs.python.org/3/library/termios.html#example. But yea, I guess they try to stay close to the C signature so someone familiar with the C/Posix side of things can look at this and say "oh yea, I know what this is doing".

qpc/cred/utils.py
```diff
@@ -13,43 +13,76 @@
 
 
 def get_password(args, req_payload, add_none=True):
```
nicolearagao, line 26: Is the suggestion below valid at all? I mean, I thought about this when I noticed that we have some code duplication that could be simplified but I am afraid that with that modification the code wouldn't be as clear to understand. There wouldn't be a clear pattern, so I am uncertain if we should take DRY as seriously here. What are your thoughts on this ?
```
def get_password(args, req_payload, add_none=True):
    """Collect the credential attribute values and place in credential dictionary.

    :param args: the command line arguments
    :param req_payload: the dictionary for the request
    :param add_none: add None for a key if True vs. not in dictionary
    :returns: the updated dictionary
    """
    req_payload = _get_one_line_auth(args, req_payload, add_none, "password", messages.CONN_PASSWORD)
    req_payload = _get_ssh_key_value(args, req_payload, add_none)
    req_payload = _get_one_line_auth(args, req_payload, add_none, "ssh_passphrase", messages.SSH_PASSPHRASE)
    req_payload = _get_one_line_auth(args, req_payload, add_none,"become_password", messages.BECOME_PASSWORD)
    req_payload = _get_token_value(args, req_payload, add_none)

    return req_payload

def _get_one_line_auth(args, req_payload, add_none, attribute, message):
    if attribute in args and getattr(args, attribute):
        with a_tty():
            print(_(message))
            pass_prompt = getpass()
            check_if_prompt_is_not_empty(pass_prompt)
            req_payload[attribute] = pass_prompt
    elif add_none:
        req_payload[attribute] = None

    return req_payload   


def _get_ssh_key_value(args, req_payload, add_none):
    """Collect the ssh_key value and place in credential dictionary."""
    if "ssh_key" in args and args.ssh_key:
        if sys.stdin.isatty():
            print(_(messages.SSH_KEY))
            pass_prompt = get_multiline_pass(prompt=f"{messages.SSH_KEY_PROMPT}")
            check_if_prompt_is_not_empty(pass_prompt)
        else:
            pass_prompt = "".join(sys.stdin.readlines())
        req_payload["ssh_key"] = pass_prompt
    elif add_none:
        req_payload["ssh_key"] = None

    return req_payload


def _get_token_value(args, req_payload, add_none):
    """Collect the token value and place in credential dictionary."""
    if getattr(args, "token", None):
        with a_tty():
            token_prompt = getpass(messages.AUTH_TOKEN)
            check_if_prompt_is_not_empty(token_prompt)
            req_payload["auth_token"] = token_prompt
    elif add_none:
        req_payload["auth_token"] = None

    return req_payload
```

infinitewarp: I think that's a good DRY idea!

bruno-fs (approved):
Thanks for the hard work =D 👏🏽

qpc/cred/utils.py
abellotti: I thought of that earlier but kept it this way for readability.  But if the general feeling is to DRY this too, I'll give it a whirl. I'll probably have the first argument as the identifier ("password", "ssh_passphrase", ...) again for readability. I'll update it shortly.

abellotti: functions dry'ed up and commit pushed. https://github.com/quipucords/qpc/pull/247/commits/2066a9a8b64683a87c444a7e935296ab8fe15819

qpc/cred/add.py
```diff
@@ -73,6 +73,12 @@ def __init__(self, subparsers):
             metavar="FILENAME",
             help=_(messages.CRED_SSH_HELP),
         )
+        group.add_argument(
+            "--sshkeyvalue",
```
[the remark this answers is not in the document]
abellotti: squashed and pushed.

nicolearagao (approved):
LGTM

PR #248, opened 2023-06-15
chore: adding setuptools as a dependency

abellotti:
chore: adding setuptools as a dependency

- Adding setuptools as a poetry dependency.
   PyCharm would not open the qpc project off the bat without it.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -694,6 +694,22 @@ files = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


bruno-fs (approved):
that's... interesting 🤔

abellotti:
Pinged you here just to make sure I'm not breaking anything downstream.

PR #258, opened 2023-07-07
Skip upload container image step on PRs from forks... for real

mirekdlugosz:
See https://github.com/quipucords/quipucords/pull/2412

This time PR is created from same repo. This is by mistake, but gives us a chance to confirm that image **will** be uploaded.


bruno-fs (approved):
it did push to quay! :+1:

ruda (approved):

PR #261, opened 2023-07-13
chore(deps): move ruff to dev dependencies group

bruno-fs:
ruff should be labeled as a dev dependency (we don't need more rust dependencies than what we already have XD)


infinitewarp (approved):
Good catch! 😅

abellotti (approved):

PR #266, opened 2023-07-24
update requirements for downstream build stuff

infinitewarp:


pyproject.toml
```diff
@@ -31,6 +31,7 @@ requests-mock = ">=1.9.3"
 pytest-mock = "^3.8.2"
 rstcheck = "^6.1.1"
 ruff = "^0.0.270"
+pip-tools = "^7.1.0"
```
bruno-fs, line 34: aren't we missing `pybuild-deps` (given its usage on `make search-build-requirements`)?

infinitewarp: Good catch! I added this as a dev dependency and amended the last commit.

I probably missed this because the command in that make target (as copied from quipucords) silences errors and always returns a clean exit code. (`2>/dev/null` ... `|| true`) ☹️

Makefile
```diff
@@ -76,3 +79,41 @@ manpage:
...
+lock-requirements: lock-main-requirements
+	rm -f requirements-build.txt requirements-build.in
+	$(MAKE) search-build-requirements
+	$(MAKE) lock-build-requirements
+	# run another rounds of search/lock for build requirements
+	# (build dependencies also have build dependencies after all :)
+	$(MAKE) search-build-requirements
+	$(MAKE) lock-build-requirements
+	$(MAKE) search-build-requirements
+	$(MAKE) lock-build-requirements
```
infinitewarp, line 92: Why is this repeated exactly twice like this? This is what the quipucords make target does, but I do not understand why it repeats.

pyproject.toml
infinitewarp: For fun, I removed the stderr redirection and `true`, and then I ran `make lock-requirements`.

It spewed out a stack trace at one point but ultimately completed with a normal `0` exit code. Here's a piece of that stack trace:

```
...
cat requirements*.txt | grep -vE '(^ )|(#)' | awk '{print $1}' | sed 's/==/ /' | \
	xargs -P6 -n2 poetry run pybuild-deps find-build-deps | \
	sort -u >> requirements-build.in
  File "/Users/brasmith/Library/Caches/pypoetry/virtualenvs/qpc-XVoVAvnv-py3.11/bin/pybuild-deps", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/brasmith/Library/Caches/pypoetry/virtualenvs/qpc-XVoVAvnv-py3.11/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
...
  File "/Users/brasmith/Library/Caches/pypoetry/virtualenvs/qpc-XVoVAvnv-py3.11/lib/python3.11/site-packages/pybuild_deps/get_package_source.py", line 24, in get_package_source
    return retrieve_and_save_source_from_pypi(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/brasmith/Library/Caches/pypoetry/virtualenvs/qpc-XVoVAvnv-py3.11/lib/python3.11/site-packages/pybuild_deps/get_package_source.py", line 37, in retrieve_and_save_source_from_pypi
    source_url = get_source_url(package_name, version)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/brasmith/Library/Caches/pypoetry/virtualenvs/qpc-XVoVAvnv-py3.11/lib/python3.11/site-packages/pybuild_deps/get_package_source.py", line 52, in get_source_url
    raise NotImplementedError()
NotImplementedError
...
```
😕 

Should we _not_ ignore this?

Makefile
bruno-fs: this search should be recursive, but I had no idea how to do it on a makefile. I added these as a guess on what seems to be a "good enough" rounds.

IMO the recursion should be something done on pybuild-deps side (integrate with pip-tools and recursively look for build dependencies + lock until no new dependencies are found)

pyproject.toml
bruno-fs: Mmmmm. I'll need to dig the blame on quipucords to see if I had any extra context for a proper answer.

bruno-fs: lol. no need to check blame... https://github.com/bruno-fs/pybuild-deps/blob/main/src/pybuild_deps/get_package_source.py#L37-L40

bruno-fs: it's an issue on `pybuild_deps` side... designed to fail like this.

It would be interesting to know which exact dependencies are causing this failure

bruno-fs: `setuptools-scm[toml]` and `celery[kombu]` are causing this (on quipucords side)

bruno-fs: replacing the `sed` command with `perl -pe 's/==/ /;s/\[\w+\]//'` suppress the errors on my system

abellotti: sed/grep/awk ftw ❤️

bruno-fs (approved):
lgtm

PR #268, opened 2023-07-24
All things pathlib

bruno-fs:
prefer `pathlib` over `os.path`.

Before merging I'll squash all 5 commits into a single one.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -29,7 +29,7 @@ QPC_PATH_CONSTANTS = (
 def mock_path_constants(tmp_path, monkeypatch):
     """Mock path constants used on qpc."""
     for path in QPC_PATH_CONSTANTS:
-        monkeypatch.setattr(f"qpc.utils.{path}", str(tmp_path / path))
+        monkeypatch.setattr(f"qpc.utils.{path}", tmp_path / path)
 
 
 def _set_path_constants_to_none():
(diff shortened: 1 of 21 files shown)


conftest.py
```diff
@@ -44,6 +44,10 @@ def pytest_collection(session):
 
     This function runs before collecting tests.
     """
+    # setting path constants to None on utils module is a dirty way to ensure
+    # no test will write on top of user config file. This is a last resort, as
+    # mock_path_constants fixture shall be doing the job of actually patching
+    # those to a usable value in tests
```
infinitewarp, line 50: Thank you for explaining this mystery function call!

qpc/cred/test_cred_add.py
```diff
@@ -36,70 +40,48 @@ def setup_class(cls):
...
+            "/bin/qpc",
+            "credential",
+            "add",
+            "--name",
+            "credential1",
+            "--username",
+            "foo",
+            "--password",
+        ]
+        with pytest.raises(SystemExit), patch.object(sys, "argv", args):
```
infinitewarp, line 61: ❤️  all these improvements around `sys.argv`

infinitewarp (approved):
So many cleanups and pythonic improvements in one PR! 😍 LGTM 👏

conftest.py
bruno-fs: almost 2 years ago me was FURIOUS after realising that running tests modified my local configuration 😅

qpc/cred/test_cred_add.py
nicolearagao: If I understood correctly, this is a better approach because it isolates the modification of sys.argv to this specific test through the context manager. So after the thest the sys.argv values are automatically restored. Is that accurate ?

nicolearagao (approved):
I learned a lot from this refactor. Thanks =)

qpc/cred/test_cred_add.py
infinitewarp: Yes, that is correct @nicolearagao! 😃

PR #269, opened 2023-07-31
1.3 fix build issues

infinitewarp:
Fixing some issues that surfaced in the downstream build.

I strongly urge reevaluating the current logic in the near future and refactoring it to be more straightforward and robust.

### DIFF
--- a/qpc/__init__.py
+++ b/qpc/__init__.py
@@ -1,6 +1,27 @@
 """QPC Package Initialization."""
 from importlib import metadata
+from pathlib import Path
+from tomllib import load
 
-# Let's get the package version from poetry
-__package__version__ = metadata.version(__package__)
+try:
+    # Let's get the package version from poetry
+    __package__version__ = metadata.version(__package__)
+except metadata.PackageNotFoundError:
+    # PackageNotFoundError may be raised if *not* called from
+    # within a poetry environment, such as during execution of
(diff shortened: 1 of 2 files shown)


qpc/__init__.py
```diff
@@ -1,11 +1,20 @@
...
 from tomllib import load
 
 try:
-    # Let's get the package version from poetry
-    __package__version__ = metadata.version(__package__)
+    # Try to get the package version from importlib.
+    # This operation can pull its information from different places depending on how
+    # the program is being called. We cannot simply use `__package__` because the
+    # package might have been given a different name at install time, which we expect
+    # to match the QPC_VAR_PROGRAM_NAME value.
```
bruno-fs, line 12: this is odd. IMO we should be changing only the script name, not the whole package name.

infinitewarp: I'm trying a different change locally that may simplify the current hack workaround fix.

infinitewarp: I yanked this change after confirming locally that we can keep `qpc` as the hard-coded package name and expect `dsc` as the program entry point if `QPC_VAR_PROGRAM_NAME=dsc` was set at build/install time.

bruno-fs (approved):

abellotti (approved):

PR #282, opened 2023-10-12
refactor: drop gitpython, run git directly to get current repo sha1

infinitewarp:
This change drops the `gitpython` third-party dependency. As discussed previously elsewhere, we need to minimize our use of third-party dependencies so that we can make an RPM for RHEL.


qpc/release.py
```diff
@@ -20,7 +19,24 @@ def get_current_sha1() -> str:
...
+        git_result = subprocess.run(
+            ("git", "rev-parse", "HEAD"),
+            env=git_env,
+            cwd=repo_root,
+            capture_output=True,
+            check=True,
+        )
+    except (FileNotFoundError, subprocess.CalledProcessError):
+        # FileNotFoundError raises when `git` program not found.
+        # CalledProcessError raises when `git` has non-zero return code.
```
abellotti, line 34: FileNotFoundError is straightforward, Should we show what the CalledProcessError was or is that detailed in a stacktrace?

infinitewarp: In both cases, I'm just swallowing the exception and not displaying anything since these are likely to be reached only when running the code from source, not the container, _and_ because this was how the original code behaved with `gitpython`. The only situation where I could normally foresee a non-zero code from git is if this code is reached when you're running it directly from source but you don't actually have the git repo with the source. All of the git-related logic here is just for us devs and QEs (mostly QEs), and specifically for when one of us is running directly from the repo, which none of our end user customers should be doing (and if they are, we don't support it).

So, basically, we're in a corner case of an internal edge case. I don't think we need to output anything else here for "running from source (unsupported!) but without the git repo (what are you doing?!)". 🙂 

FWIW, this function will likely change yet again after we go to RPM releases since we'll have neither the git repo nor the environment variable. As we get into building RPMs, if our QEs will stop installing from source entirely, thereby obviating the need for _any_ of this git code, we may just drop all of this. (The original motivation for dropping `gitpython` is that it's not available as a standard RHEL RPM, last we checked.)

abellotti: Gotcha, if it's for `us` we should know what we're doing 😉 

Thank you for the explanation.

abellotti (approved):

bruno-fs (approved):
LGTM

Not work for this PR but I do think we should not rely on env var for the upcoming RPM, tho. Maybe we should write to a file that will be in a specific place relative to project root that can be generated during build time?

PR #285, opened 2023-10-24
replace jinja-cli with local python script

infinitewarp:
Packages for `jinja-cli` and `jinja-cli2` are not available in our downstream build system. Our use case for rendering Jinja template content is very simple, though. So, I wrote this small script to use the `jinja2` library directly (which we _do_ have available downstream for both RHEL8 and RHEL9) to replace our single use of the `jinja` program, and I dropped `jinja-cli` from our Poetry environment and replaced it with just `jinja2`.

I deliberately made this a standalone script and _not_ part of the `qpc` package because we want to run this as a build step downstream, and we don't need to install all of qpc just to generate the man page.


bruno-fs (approved):
LGTM

nicolearagao (approved):

PR #286, opened 2023-10-25
fix: warning: Selection of nursery rule `PLC1901` without the `--preview` flag is deprecated.

infinitewarp:
This should suppress the warnings that `ruff` has recently started outputting:

```
warning: Selection of nursery rule `PLC1901` without the `--preview` flag is deprecated.
warning: Selection of nursery rule `PLC1901` without the `--preview` flag is deprecated.
warning: Selection of nursery rule `PLC1901` without the `--preview` flag is deprecated.
```

Reference: https://docs.astral.sh/ruff/preview/#using-rules-that-are-in-preview

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,8 @@ ignore = [
 ]
 src = ["qpc"]
 target-version = "py311"
+preview = true
+explicit-preview-rules = true
 
 [tool.ruff.per-file-ignores]
 "**/commands.py" = ["F401"]


bruno-fs:
LGTM, but before we merge this would u mind checking it updating ruff can fix this? (we might need to bump it manually because it went from 0.0 to 0.1 (`poetry add -G dev ruff@latest`)

infinitewarp:
Closing because https://github.com/quipucords/qpc/pull/287 also resolves this issue.

PR #301, opened 2023-12-04
use packit to build upstream RPMs

infinitewarp:


.packit.yaml
```diff
@@ -0,0 +1,36 @@
...
+    trigger: pull_request
+    targets: &targets_list
+      # we officially support rhel-8 and rhel-9 downstream.
+      - rhel-8
+      - rhel-9
+      # centos-stream is like a preview of upcoming rhel.
+      - centos-stream-8
+      - centos-stream-9
+      # fedora is like an even farther future preview.
+      - fedora-37
```
bruno-fs, line 19: I get we need fedora 38 for Jenkins, but why fedora 37? 🪓

infinitewarp: Good point! I only included that because it was part of Packit's `fedora-all` group alias that targets all current supported versions. 🪓-ing!

qpc.spec
```diff
@@ -0,0 +1,54 @@
...
+%license LICENSE
+%doc README.md
+%doc %{_mandir}/man1/qpc.*
+%{_bindir}/qpc
+%{python3_sitelib}/qpc/
+%{python3_sitelib}/qpc-*.egg-info/
+
+%changelog
+* Thu Oct 12 2023 Brad Smith <brasmith@redhat.com> - 0:1.5.0a1-1
+- Initial prototype RPM for upcoming 1.5.0 release series.
```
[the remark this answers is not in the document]
infinitewarp: I wrote this one by hand. There may be tools to assist (I'm _pretty sure_ the Packit CLI can help here), but I have not researched or tested deliberately.

PR #304, opened 2023-12-18
vendor `packaging` due to issues with installing it in some downstream environments

infinitewarp:
`packaging` is distributed under the two-clause BSD license and Apache 2.0, and I believe it should be compatible to include in our GPLv3-licensed project, but I am not a lawyer. gnu.org [claims two-clause BSD is compatible with GPL](https://www.gnu.org/licenses/bsd.html) (see the very end of the document) and Apache.org [claims Apache 2.0 is compatible with GPLv3](https://www.apache.org/licenses/GPL-compatibility.html).

To perform the vendoring operation, I added `vendorize` as a dev dependency. To use it, simply execute its command from the root directory of the repo.

```bash
poetry run python-vendorize
```

This enables us to drop the sometimes problematic `python%{python3_pkgversion}-packaging` from our spec files, but it incurs the cost of growing our project's footprint and attack surface.


bruno-fs:
Given our [very limited case of packaging](https://github.com/quipucords/qpc/blob/caa75c18e0d2e6dd670a4a77ace1a67e0ee1163d/qpc/request.py#L52-L59), I think
 @mirekdlugosz suggestion to drop it in favor of a naive ad hoc implementation is very feasible and easy to maintain.
 
 Here's a suggestion
 
 ```python
 def version_tuple(version_str):
    version_pattern = r"(\d+)\.(\d+)\.(\d+)"
    res = re.match(version_pattern, version_str)
    return tuple(map(int, res.groups(0)))

a = "1.2.3"
b = "1.12.1"
c = "1.2.3+some-sha"

assert not version_tuple(a) > version_tuple(b)
assert version_tuple(b) > version_tuple(a)
assert version_tuple(a) == version_tuple(c)  # for our purposes I think it is fine to consider a "dev" version the same as a "production" one.
 ```

PR #305, opened 2023-12-19
refactor: use custom function to parse version strings for comparison

infinitewarp:
Followup to https://github.com/quipucords/qpc/pull/304#issuecomment-1862806266

Important note: this simplified helper function **does not** support all of the standard version patterns defined in [PEP-440](https://peps.python.org/pep-0440/). The qpc dev team agreed that we will support only this simplified pattern, and we will either always release exclusively using `X.Y.Z` integers (e.g. `1.2.3`) or expect "fuzzy" matching to discard anything after the third integer segment (e.g. `1.2.3a1` is treated equal to `1.2.3`).


requirements-build.txt
```diff
@@ -34,14 +34,15 @@ semantic-version==2.10.0
     # via setuptools-rust
 setuptools-rust==1.8.1
     # via cryptography
-setuptools-scm==8.0.4
+setuptools-scm @ https://files.pythonhosted.org/packages/0e/a3/b9a8b0adfe672bf0df5901707aa929d30a97ee390ba651910186776746d2/setuptools_scm-8.0.4-py3-none-any.whl#sha256=b47844cd2a84b83b3187a5782c71128c28b4c94cad8bfb871da2784a5cb54c4f
```
bruno-fs, line 37: oO
is this.. right? I might need to check if `pybuild-deps` has a bug producing this

bruno-fs (commented):
Pretty good! I'm only concerned with an (automatic?) change on setuptools-scm build dependency. I'm almost sure this new requirements-build.txt won't be accepted by cachito

requirements-build.txt
infinitewarp: 🤷 I didn't touch this directly. I just let our lock command do its thing. I'll look into it too…

bruno-fs: this is so weird... setuptools-scm is the exact same version 🤔 
might be related to line 42 pointing that it depends on itself

bruno-fs (approved):

PR #306, opened 2023-12-19
Drop container requirements

infinitewarp:
Now that we are building and distributing RPMs (or, at least, we are _ready_ to do so), it is time to stop building qpc containers.

This also includes a version bump to `1.5.0` so we don't accidentally include this change in any other builds or containers that use `qpc` `1.4.x`.


bruno-fs (approved):
:shipit:

PR #314, opened 2024-02-14
test: port CLI tests from Camayoc

mirekdlugosz:
A little unexpectedly, there were few tests in Camayoc that checked CLI input validation. They bail out before any requests are sent, so can be safely tested at the unit level.

### DIFF
--- a/qpc/tests/scan/test_scan_add.py
+++ b/qpc/tests/scan/test_scan_add.py
@@ -178,10 +178,10 @@ class TestScanAddCli:
         "disable_opt_products_value", ("1", "-100", "redhat_packages", "ifconfig")
     )
     def test_disable_optional_products_negative(
-        self, capsys, disable_opt_products_value
+        self, capsys, mocker, disable_opt_products_value
     ):
         """Test add scan with unknown disabled-optional-products value."""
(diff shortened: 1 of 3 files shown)


qpc/tests/scan/test_scan_add.py
```diff
@@ -174,6 +174,32 @@ def test_disable_optional_products(self, caplog):
...
+            "/bin/qpc",
+            "scan",
+            "add",
+            "--name",
+            "scan_1",
+            "--sources",
+            "source_1",
+            "--disabled-optional-products",
+            disable_opt_products_value,
+        ]
```
bruno-fs, line 194: I know MOST tests are patching `sys.argv` directly but we are preferring to mock it instead, like here
https://github.com/quipucords/qpc/blob/87bd654f83ee5ae85a3b4701322e1c43589dee78/docs/test_jina_render.py#L95

mirekdlugosz: Fixed in https://github.com/quipucords/qpc/pull/314/commits/38f9a58c6f6c68b6b06462c54681e0eaa5170470 . We can squash during merge, added a commit for easier review.

PR #323, opened 2024-04-05
fix: delay a second when paging results

ruda:


bruno-fs (changes requested):
We can hardly call this a fix.

Won't [Alberto PR](https://github.com/quipucords/quipucords/pull/2617) solve this particular pain for camayoc? or is camayoc handling with 100+ items?

If not, we should find a proper solution. Not this.

ruda:
> We can't hardly call this a fix.
> 
> Won't [Alberto PR](https://github.com/quipucords/quipucords/pull/2617) solve this particular pain for camayoc? or is camayoc handling with 100+ items?
> 
> If not, we should find a proper solution. Not this.

OK.

PR #328, opened 2024-04-30
style: drop `black` in favor of `ruff format`

infinitewarp:


qpc/tests/cred/conftest.py
```diff
@@ -4,7 +4,8 @@
 
 
 @pytest.fixture(autouse=True)
-def _setup_server_config_file(server_config): ...
+def _setup_server_config_file(server_config):
+    ...
```
bruno-fs, line 8: mmm I liked the oneliner style black had for these. but not enough to keep this from merging  🤷🏽 😆

bruno-fs (approved):

PR #333, opened 2024-05-22
chore: Adding support for Fedora 39

abellotti:
- Updated the packit config yaml to include Fedora 39 builds for both x86_64 and ARM architectures.
- Bumped poetry and pyproject to use 3.12


abellotti:
Closing this, preferring #337 to also include Fedora 40.

PR #337, opened 2024-05-22
chore: Adding support for Fedora 39 and 40

abellotti:
- Updated the packit config yaml to include Fedora 39 and 40 builds for both x86_64 and ARM architectures.
- Dropped CentOS Stream 8 (baseOS image was removed from Packit)
- Bumped poetry and pyproject to use 3.12
- Pulling in setuptools 70.0.0


pyproject.toml
```diff
@@ -1,12 +1,18 @@
 [project]
-requires-python = ">=3.11, <3.13"
+name = "qpc"
+version = "1.8.0"
```
infinitewarp, line 3: Do we need this `version` specification? This adds another place to remember to update every time we bump versions to release. Or can we remove the `version` down under `[tool.poetry]` if we prefer/need this one under `[project]`?

Ditto for `name` since we have to patch that downstream.

abellotti: poetry vs 3.12 setuptools fighting each other :( 

with name and version only in `[project]`:

```
$ poetry run qpc

The Poetry configuration is invalid:
  - The fields ['name', 'version'] are required in package mode.
```

with name and version only in `[tool.poetry]`:

```
$ python setup.py
...
ValueError: invalid pyproject.toml config: `project`.
configuration error: `project` must contain ['name'] properties
```

bruno-fs: I don't know by heart, but I know specfiles in RHEL do support PEP-517/518 compliant projects (aka, without setup.py). Maybe this is a good time to get rid of `setup.py` for good?

infinitewarp: Understood. I was afraid that might happen. 😞 

In that case, we should add a check somewhere to ensure that these two fields stay in sync. Maybe add a make target that we can run locally and in a GitHub action to ensure that we never update one and forget the other. We could use `tomllib` to read the file and check these two fields, assuming there's not a better way to parse it.

bruno-fs: Only thing holding us back is that we would also need to change downstream patching mechanism (maybe we could also just get rid of qpc name and be done with it)

infinitewarp: Orrrrr maybe we consider what @bruno-fs said. 🤔 

(Beat me by seconds?? gotta go fast. 🦔)

bruno-fs: > In that case, we should add a check somewhere to ensure that these two fields stay in sync.

I think I did that in chaski with some third party tool 🤔... lemme check

bruno-fs: Found it.
https://github.com/quipucords/chaski/blob/main/pyproject.toml#L31C1-L34C1

the tool accepts pointing python files as well

https://commitizen-tools.github.io/commitizen/commands/bump/#version_files

bruno-fs: Doing a quick dig, I found these macros

```
%generate_buildrequires
%pyproject_buildrequires
%pyproject_wheel
%install
%pyproject_install
%pyproject_save_files
```
source: https://src.fedoraproject.org/rpms/pyproject-rpm-macros

However, if you decide to go this path, I'd invert the order of implementation and try to make it work downstream FIRST as the availability of dependencies is more scarce there.

TIP: we **_DON'T NEED POETRY_**. Just `poetry-core` as build dependency should suffice (if the macro does not "magically" solve it)

bruno-fs: example implementation: https://url.corp.redhat.com/13ecab7
(cryptography doesn't rely on poetry but it is a package that doesn't have setup.py)

bruno-fs: Just to keep everyone on the loop, in order to use pyproject template tags (and poetry-core) on RHEL8 we would need to build them all from scratch. Let's postpone removing setup.py as this "tech debt" is clearly big enough to deserve its own Jira ticket.

config-verify.py
```diff
@@ -0,0 +1,29 @@
+"""QPC pyproject toml file verification."""
+
+import sys
+from pathlib import Path
+from tomllib import load
```
bruno-fs, line 5: stdlib ftw, ALWAYS <3

bruno-fs (approved):
LGTM

ruda (approved):

PR #341, opened 2024-06-05
Aggregate report: Add support for --output-file flag

mirekdlugosz:
Add support for optional `--output-file` flag to `qpc report aggregate`, to maintain consistency with `qpc report deployments` and `qpc report details`.


qpc/report/aggregate.py
```diff
@@ -44,11 +48,24 @@ def __init__(self, subparsers):
...
+            metavar="PATH",
+            help=_(messages.REPORT_PATH_HELP),
+        )
         self.report_id = None
 
     def _validate_args(self):
         CliCommand._validate_args(self)
 
+        try:
+            if (args_path := getattr(self.args, "path", None)) is not None:
```
mirekdlugosz, line 63: [details are doing `if self.args.path`](https://github.com/quipucords/qpc/blob/main/qpc/report/details.py#L91). That has two consequences:

- they need to set `path` in `Namespace()` in all unit tests
- if `self.args.path` is empty string, it will only bail out after making a request, in `write_file` call

I decided I don't want to modify existing unit tests for aggregate (after all, `--output-file` is supposed to be optional), and I'd rather catch empty value early than during `Path().open()` call.

infinitewarp (approved):

PR #343, opened 2024-06-17
Minor cleanup

mirekdlugosz:
I wanted to work on DISCOVERY-458 , but then I discovered that there's hardly anything to do. But since I was there already, I decided to clean up things a little.

f13d2ef8dec8fb6a9dfa27defc54a92046b39102 removes `filehandle.close()` call after context manager. File is closed by context manager on exit.

a94d8f26f45b5f4fb7695c30c8e06b2e81fd55e6 removes a bunch of variables from `messages.py`. These variables are no longer used - they generally belong to features that have since been removed (old insights upload, restart/pause, separate merge and upload), but some of them date back to when CLI and backend shared single repo - they were never really used in CLI.

### DIFF
--- a/qpc/insights/publish.py
+++ b/qpc/insights/publish.py
@@ -145,7 +145,6 @@ class InsightsPublishCommand(CliCommand):
             successfully_submitted = self._make_publish_request(
                 insights_client, insights.INGRESS_REPORT_URI, files
             )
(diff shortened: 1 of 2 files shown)


infinitewarp (approved):

nicolearagao (approved):
LGTM

PR #352, opened 2024-10-15
feat: use new credentials v2 api

infinitewarp:
See https://github.com/quipucords/quipucords/pull/2744 for details about the new API.

Relates to JIRA: DISCOVERY-797

### DIFF
--- a/qpc/tests/cred/test_openshift_cred_add.py
+++ b/qpc/tests/cred/test_openshift_cred_add.py
@@ -91,10 +91,7 @@ class TestOpenShiftAddCredential:
             CLI().main()
         out, err = capsys.readouterr()
         assert out == ""
-        assert (
-            "one of the arguments --password --sshkeyfile"
-            " --sshkey --token is required" in err
-        )
+        assert "one of the arguments --password --sshkey --token is required" in err
 
     def test_add_no_name(
         self,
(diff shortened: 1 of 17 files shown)


infinitewarp:
I am changing this PR to draft status while we figure out some integration test changes in camayoc.

mirekdlugosz:
In `qpc/cred/{add,edit}.py` we should remove `--sshkeyfile` flag, and possibly most / all references to `filename` arg variable in `qpc/cred/utils.py`, `qpc/tests/cred/test_cred_add.py` and `qpc/tests/cred/test_cred_edit.py`, possibly other places as well.

`ssh_keyfile` key in payload is marked as `read_only` in v2, so it doesn't make sense to send it, and we should not expose it as an available option.

qpc/tests/cred/test_openshift_cred_add.py
```diff
@@ -91,10 +91,7 @@ def test_add_no_token(
             CLI().main()
         out, err = capsys.readouterr()
         assert out == ""
-        assert (
-            "one of the arguments --password --sshkeyfile"
-            " --sshkey --token is required" in err
-        )
+        assert "one of the arguments --password" " --sshkey --token is required" in err
```
mirekdlugosz, line 94:
```suggestion
        assert "one of the arguments --password --sshkey --token is required" in err
```

I think black tends to no join strings where they used to be on multiple lines. Not that it really matters.

infinitewarp: Good catch. This happened as a result of `ruff` autoformatting the file, and I missed it. I'll fix this manually.

infinitewarp:
The [continuous-integration/jenkins/pr-head](https://stage-jenkins-csb-smqe.apps.ocp-c1.prod.psi.redhat.com/job/Discovery%20upstream%20PR%20-%20qpc/job/PR-352/9/display/redirect) check fails because `camayoc` requires additional changes that correspond with this PR's changes. Mirek ran [standalone job 224](https://stage-jenkins-csb-smqe.apps.ocp-c1.prod.psi.redhat.com/view/Discovery/job/discovery-standalone/224/) with the two new branches, and it completed successfully.

This branch must merge at the same time or very shortly after https://github.com/quipucords/camayoc/pull/554 merges.

PR #358, opened 2025-02-03
build: bump qpc version to 1.12.0, minimum server version to 1.12.0, update requirements

infinitewarp:


infinitewarp:
I expect `continuous-integration/jenkins/pr-head` to fail at first because this change requires server version 1.12.0, but that version is not yet merged and available.

infinitewarp:
As expected, `continuous-integration/jenkins/pr-head` did fail the first time, but after `1.12` server became available and I reran the job, it passed.

PR #363, opened 2025-03-24
fix: Don't warn about non-existing config file when creating it

mirekdlugosz:
This fixes DISCOVERY-815 .

Before this patch, when you run `dsc server config` for the first time ever, you get notification about config file not existing. That is triggered by `read_require_auth()` call. The easiest way is indenting this entire block - so when command is `server config`, we go straight to running a command. For anything else, we keep the current logic.

I've run all Camayoc cli tests using qpc with that patch and they passed.

### DIFF
--- a/qpc/cli.py
+++ b/qpc/cli.py
@@ -56,7 +56,6 @@ from qpc.utils import (
     get_server_location,
     logger,
     read_client_token,
-    read_require_auth,
     setup_logging,
 )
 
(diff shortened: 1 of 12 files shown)


qpc/cli.py
```diff
@@ -218,10 +218,10 @@ def main(self):
                 logger.error(_(messages.SERVER_CONFIG_REQUIRED), QPC_VAR_PROGRAM_NAME)
                 sys.exit(1)
 
-        if read_require_auth():
-            if (not is_server_cmd or is_server_logout) and not read_client_token():
-                logger.error(_(messages.SERVER_LOGIN_REQUIRED), QPC_VAR_PROGRAM_NAME)
-                sys.exit(1)
+            if read_require_auth():
```
mirekdlugosz, line 221: So, um, this is effectively `if True`.

`read_require_auth()` will return `True` if config file does not exist.

If it exists, it reads `require_token` from file. But that one is `True` by default and **there is no code in qpc to change it**. So the only way this function would ever return `False` is when someone manually modifies the config file and switches the value.

It's also very unclear to me why that setting even exists. It doesn't make sense to store that globally. Authentication is required at endpoint level, so each command should be able to answer if it requires authentication or not.

But maybe the idea was to store information if user ever logged in successfully. If so, that logic is long gone, or was never implemented. Not to mention that you can very much answer that based on `client_token` file presence and content.

qpc/cli.py
```diff
@@ -218,10 +218,10 @@ def main(self):
...
+                if (not is_server_cmd or is_server_logout) and not read_client_token():
```
mirekdlugosz, line 222: So if I already did `qpc server config` and try `qpc server logout`, it will tell me that I need to log in first :upside_down_face: 

I guess it's not wrong, but if we think about commands as ways to achieve a certain state, then that state is already achieved and there's no reason to warn the user.

bruno-fs (approved):

infinitewarp (approved):

infinitewarp (approved):

PR #364, opened 2025-03-24
fix: Tell user when insights report can't be generated due to source types

mirekdlugosz:
All examples assume user is interacting with report that does not support insights report (so scan used sources of type {openshift, rhacs, ansible}).

Before the patch:

```
$ qpc -v report insights --report 2
Insights report 2 does not exist.
```

```
$ qpc -v report insights --scan-job 5
No Insights report available for scan job 5.
```

After this patch:

```
$ qpc -v report insights --report 2
Insights report 2 does not exist. Insights report can be generated only if at least one source type is network, satellite or vcenter.
```

```
$ qpc -v report insights --scan-job 5
No Insights report available for scan job 5. Insights report can be generated only if at least one source type is network, satellite or vcenter.
```

If user tries to download report by using non-existing report id or scan job id, clarification message is **not** shown (so, this patch does not change that case). 

There are two possible problems:

* When using a server **without** https://github.com/quipucords/quipucords/pull/2866 , API request to `/api/v2/jobs/?report_id=ID` will return **all** scanjobs. qpc will then make a decision based on whatever happens to be first item on the list. I assume we can guard against this with `min_server_version` when making a release and we don't need more extensive handling on qpc side.
* Removing a scan does not remove a report, so it is possible to request insights report for report without scanjobs, in which case we can't tell what source types were originally used. We still try to download it, we just don't show clarification in case of a problem. Since qpc does not provide a way to list reports, only scans and scanjobs, I don't think that's a large problem.

Relates to JIRA: DISCOVERY-791

### DIFF
--- a/qpc/messages.py
+++ b/qpc/messages.py
@@ -209,6 +209,10 @@ REPORT_NO_DETAIL_REPORT_FOR_SJ = "No report detail available for scan job %s."
 REPORT_NO_DETAIL_REPORT_FOR_REPORT_ID = "No report detail available for report id %s."
 REPORT_NO_INSIGHTS_REPORT_FOR_SJ = "No Insights report available for scan job %s."
 REPORT_NO_INSIGHTS_REPORT_FOR_REPORT_ID = "Insights report %s does not exist."
+REPORT_NO_INSIGHTS_CLARIFICATION = (
+    "Insights report can be generated only if at least one source type "
+    "is network, satellite or vcenter."
+)
 REPORT_OUTPUT_CANNOT_BE_EMPTY = "%s cannot be empty string."
 REPORT_OUTPUT_IS_A_DIRECTORY = "%s %s was a directory."
 REPORT_DIRECTORY_DOES_NOT_EXIST = "The directory %s does not exist.  Cannot write here."
--- a/qpc/messages.py
+++ b/qpc/messages.py
@@ -211,7 +211,7 @@ REPORT_NO_INSIGHTS_REPORT_FOR_SJ = "No Insights report available for scan job %s
 REPORT_NO_INSIGHTS_REPORT_FOR_REPORT_ID = "Insights report %s does not exist."
 REPORT_NO_INSIGHTS_CLARIFICATION = (
     "Insights report can be generated only if at least one source type "
-    "is network, satellite or vcenter."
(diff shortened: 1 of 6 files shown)


qpc/scan/__init__.py
```diff
@@ -23,6 +23,7 @@
 SCAN_URI = "/api/v1/scans/"
 SCAN_BULK_DELETE_URI = "/api/v1/scans/bulk_delete/"
 SCAN_JOB_URI = "/api/v1/jobs/"
+SCAN_JOB_V2_URI = "/api/v2/jobs/"
```
mirekdlugosz, line 26: Unfortunately I have to use API v2 for scanjobs, as v1 does not support listing all scanjobs.

At the same time, I'm not sure if we want to port qpc to scanjobs v2 **now**.

bruno-fs: I think it is fine to migrate now

qpc/messages.py
```diff
@@ -209,6 +209,10 @@
 REPORT_NO_DETAIL_REPORT_FOR_REPORT_ID = "No report detail available for report id %s."
 REPORT_NO_INSIGHTS_REPORT_FOR_SJ = "No Insights report available for scan job %s."
 REPORT_NO_INSIGHTS_REPORT_FOR_REPORT_ID = "Insights report %s does not exist."
+REPORT_NO_INSIGHTS_CLARIFICATION = (
+    "Insights report can be generated only if at least one source type "
+    "is network, satellite or vcenter."
```
infinitewarp, line 214: Very minor nitpick. (Oxford comma)
```suggestion
    "is network, satellite, or vcenter."
```

qpc/scan/__init__.py
mirekdlugosz: As discussed on Slack, API v2 does not provide endpoint for canceling running scans.

In f21148c811f625ccdd30837da34012dfd5fe6026 I made v2 default for scanjobs, while leaving canceling on v1.

bruno-fs (approved):

infinitewarp (approved):
I approve of this v1/v2 compromise! 🙂

nicolearagao (approved):

ruda (approved):

PR #366, opened 2025-04-01
build: replace poetry with uv

infinitewarp:
See rationale for migration in:

* https://github.com/quipucords/quipucords/pull/2775
* https://github.com/quipucords/quipucords/commit/ee0570641ed45fcf8544d0433d0dd9732f9b96cd


pyproject.toml
```diff
@@ -1,54 +1,60 @@
...
+    "pybuild-deps>=0.1.1,<0.2",
 ]
-authors = ["QPC Team <quipucords@redhat.com>"]
+build = ["sphinx>=7.2.6,<8"]
+
+[tool.uv]
+default-groups = [
+    "dev",
+    "build",
+]
```
abellotti, line 39: maybe add `required-version = ">=0.6.11,<0.7"` for consistency with main quipucords.

Makefile
```diff
@@ -1,6 +1,6 @@
 DATE		= $(shell date)
-PYTHON		= $(shell poetry run which python 2>/dev/null || which python)
-PKG_VERSION = $(shell poetry -s version)
+PYTHON		= $(shell uv run which python 2>/dev/null || which python)
+PKG_VERSION = $(shell uv run python get-version.py)
```
bruno-fs, line 3: optional, since we have `yq` installed for some quipucords make targets

```suggestion
PKG_VERSION = $(shell yq '.project.version' pyproject.toml)
```

config-verify.py
```diff
@@ -16,23 +16,11 @@
...
 
-    if project["name"] != tool_poetry["name"]:
+    if spec_version != project["version"]:
         print(
-            "Project name {n1} does not match tool.poetry name {n2}".format(
-                n1=project["name"], n2=tool_poetry["name"]
-            ),
-            file=sys.stderr,
-        )
-        sys.exit(1)
```
bruno-fs, line 28: THANK GOD WE DON'T NEED THIS ANYMORE 🙏🏽

config-verify.py
```diff
@@ -16,23 +16,11 @@
...
-
-    if spec_version != project["version"] or spec_version != tool_poetry["version"]:
-        print(
-            "Versions {v1} (project) and {v2} (tool.poetry) in pyproject.toml, and "
-            "{v3} in qpc.spec are inconsistent.".format(
-                v1=project["version"], v2=tool_poetry["version"], v3=spec_version
```
bruno-fs, line 34: 🪓

pyproject.toml
infinitewarp: Good catch! Added in a fixup commit that I will squash assuming CI checks are still happy.

abellotti (approved):

Makefile
infinitewarp: I thought about this too, but I didn't feel like declaring `yq` elsewhere like the readme and the github workers because using `yq` here would make it required for _anything_ that uses _any_ `make` commands, whereas with quipucords I think `yq` is only really needed for the Konflux pipeline. Maybe later if/when Konflux starts building qpc… 😉

PR #372, opened 2025-05-05
Add support for the --sshkeyfile parameter

abellotti:
- Support the --sshkeyfile argument for allowing users to input the ssh key by filename.  Allow for - as filename to take stdin as it would behave today for --sshkey
 - Drop support for the --sshkey argument.

Draft for now ...

to do's:
- [x] additional tests
- [x] code coverage
- [x] man page update

Solves: [DISCOVERY-811](https://issues.redhat.com/browse/DISCOVERY-811)
Requires: [camayoc/582](https://github.com/quipucords/camayoc/pull/582)

### DIFF
--- a/qpc/cred/add.py
+++ b/qpc/cred/add.py
@@ -68,10 +68,10 @@ class CredAddCommand(CliCommand):
             help=_(messages.CRED_PWD_HELP),
         )
         group.add_argument(
-            "--sshkey",
-            dest="ssh_key",
-            action="store_true",
-            help=_(messages.CRED_SSH_KEY_HELP),
+            "--sshkeyfile",
+            dest="ssh_keyfile",
+            metavar="SSH_KEYFILE",
+            help=_(messages.CRED_SSH_KEYFILE_HELP),
         )
         group.add_argument(
             "--token",
--- a/qpc/cred/add.py
+++ b/qpc/cred/add.py
@@ -83,7 +83,7 @@ class CredAddCommand(CliCommand):
             "--sshpassphrase",
             dest="ssh_passphrase",
             action="store_true",
-            help=_(messages.CRED_SSH_PSPH_HELP),
+            help=_(messages.CRED_SSH_PASSPHRASE_HELP),
         )
         self.parser.add_argument(
             "--become-method",
--- a/qpc/cred/edit.py
+++ b/qpc/cred/edit.py
@@ -73,7 +73,7 @@ class CredEditCommand(CliCommand):
             "--sshpassphrase",
             dest="ssh_passphrase",
             action="store_true",
-            help=_(messages.CRED_SSH_PSPH_HELP),
+            help=_(messages.CRED_SSH_PASSPHRASE_HELP),
         )
         self.parser.add_argument(
             "--become-method",
--- a/qpc/cred/edit.py
+++ b/qpc/cred/edit.py
@@ -64,10 +64,10 @@ class CredEditCommand(CliCommand):
             help=_(messages.CRED_TOKEN_HELP),
         )
         group.add_argument(
-            "--sshkey",
-            dest="ssh_key",
-            action="store_true",
-            help=_(messages.CRED_SSH_KEY_HELP),
+            "--sshkeyfile",
+            dest="ssh_keyfile",
+            metavar="SSH_KEYFILE",
+            help=_(messages.CRED_SSH_KEYFILE_HELP),
         )
         self.parser.add_argument(
             "--sshpassphrase",
@@ -102,7 +102,7 @@ class CredEditCommand(CliCommand):
         if not (
             self.args.username
             or self.args.password
-            or self.args.ssh_key
+            or self.args.ssh_keyfile
             or self.args.ssh_passphrase
             or self.args.become_method
             or self.args.become_user
--- a/qpc/cred/utils.py
+++ b/qpc/cred/utils.py
@@ -64,27 +64,33 @@ def _get_attribute_value(attribute, prompt, args, req_payload, add_none):
 def _get_ssh_key_value(args, req_payload, add_none):
     """Collect the ssh_key value and place in credential dictionary."""
     if "ssh_keyfile" in args and args.ssh_keyfile:
-        if args.ssh_keyfile == "-":
-            if sys.stdin.isatty():
-                print(_(messages.SSH_KEY))
-                pass_prompt = get_multiline_pass(prompt=f"{messages.SSH_KEY_PROMPT}")
-                check_if_prompt_is_not_empty(pass_prompt)
+        try:
+            if args.ssh_keyfile == "-":
+                if sys.stdin.isatty():
+                    print(_(messages.SSH_KEY))
+                    pass_prompt = get_multiline_pass(
+                        prompt=f"{messages.SSH_KEY_PROMPT}"
+                    )
+                    check_if_prompt_is_not_empty(pass_prompt)
+                else:
+                    pass_prompt = "".join(sys.stdin.readlines())
+                req_payload["ssh_key"] = pass_prompt
             else:
-                pass_prompt = "".join(sys.stdin.readlines())
-            req_payload["ssh_key"] = pass_prompt
-        else:
-            try:
-                req_payload["ssh_key"] = Path(args.ssh_keyfile).read_text()
-            except FileNotFoundError:
-                logger.error(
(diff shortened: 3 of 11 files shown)


mirekdlugosz:
Are we OK with bringing back `--sshkeyfile` flag, with different semantics, and dropping `--sshkey`? We just removed `--sshkeyfile` in December, and `--sshkey` was added in July 2023 (but I think it wasn't really advertised until much later).

I'm not saying it's wrong, and in context of CLI usage it probably doesn't really matter, but I feel changes to public interface like that deserve a proper discussion. FWIW jira ticket does not cover flag changes.

---------------------

There's inconsistency in how various flags handle prompting and stdin from pipe.

You can pipe file content to `--sshkeyfile`:

```
cat ~/.ssh/mykey | qpc cred add --type network --name mycred --username cloud-user --sshkeyfile -
```

But you **can't** pipe file content to passphrase and others:

```
$ cat ~/mypassword | qpc cred add --type network --name mycred --username cloud-user --sshkeyfile ~/.ssh/mykey --sshpassphrase
Cannot prompt for password, sshpassphrase, become_password or auth_token if not connected to a tty.
```

Trying to give `-` as argument does not help:

```
$ cat ~/mypassword | qpc cred add --type network --name mycred --username cloud-user --sshkeyfile ~/.ssh/mykey --sshpassphrase -
qpc: error: unrecognized arguments: -
```

If we decide to fix that by teaching `--sshpassphrase` to read from pipe, we need to still fail when multiple flags try to read from the same pipe (as we can't differentiate what should go where - or can we?):

```
$ cat ~/.ssh/mykey | qpc cred add --type network --name mycred --username cloud-user --sshkeyfile - --sshpassphrase
Cannot prompt for password, sshpassphrase, become_password or auth_token if not connected to a tty.
```

abellotti:
yea, not changing --sshpassphrase behavior here, so allowing that only if prompted for on a tty, so either if prompting for the SSH private key or reading the SSH private key from file, we cannot pass both on stdin.

Without passphrase:

```
$ uv run python qpc cred add --type network --username abellotti --name aab_key_from_file_nopassphrase  --sshkeyfile $HOME/.ssh/id_keycloak_arm

$  cat $HOME/.ssh/id_keycloak_arm | uv run python qpc cred add --type network --username abellotti --name aab_key_from_stdin_nopassphrase --sshkeyfile -
```

With passphrase:

```
$ uv run python qpc cred add --type network --username abellotti --name aab_key_from_file_passphrase  --sshkeyfile $HOME/.ssh/id_keycloak_arm --sshpassphrase
Provide a passphrase for the SSH Key.
Password: 

$ uv run python qpc cred add --type network --username abellotti --name aab_key_from_tty_passphrase --sshkeyfile - --sshpassphrase
Provide a Private SSH Key followed by a Control-D on a new-line.
Private SSH Key: 
Provide a passphrase for the SSH Key.
Password:
```

and the current error check as you noticed if requiring sshpassphrase not on a tty.

```
$ cat $HOME/.ssh/id_keycloak_arm | uv run python qpc cred add --type network --username abellotti --name aab_key_from_stdin_passphrase --sshkeyfile - --sshpassphrase
Cannot prompt for password, sshpassphrase, become_password or auth_token if not connected to a tty.
```

qpc/cred/edit.py
```diff
@@ -64,16 +64,16 @@ def __init__(self, subparsers):
             help=_(messages.CRED_TOKEN_HELP),
         )
         group.add_argument(
-            "--sshkey",
-            dest="ssh_key",
-            action="store_true",
-            help=_(messages.CRED_SSH_KEY_HELP),
+            "--sshkeyfile",
+            dest="ssh_keyfile",
+            metavar="SSH_KEYFILE",
```
infinitewarp, line 69: What is `metavar` doing here?

qpc/cred/add.py
```diff
@@ -68,10 +68,10 @@ def __init__(self, subparsers):
             help=_(messages.CRED_PWD_HELP),
         )
         group.add_argument(
-            "--sshkey",
-            dest="ssh_key",
-            action="store_true",
-            help=_(messages.CRED_SSH_KEY_HELP),
+            "--sshkeyfile",
+            dest="ssh_keyfile",
+            metavar="SSH_KEYFILE",
```
infinitewarp, line 73: What is `metavar` doing here?

qpc/cred/utils.py
```diff
@@ -62,14 +63,29 @@ def _get_attribute_value(attribute, prompt, args, req_payload, add_none):
...
-            pass_prompt = "".join(sys.stdin.readlines())
-        req_payload["ssh_key"] = pass_prompt
+            try:
+                req_payload["ssh_key"] = Path(args.ssh_keyfile).read_text()
+            except FileNotFoundError:
+                logger.error(
+                    _(messages.CRED_SSH_KEYFILE_DOES_NOT_EXIST), args.ssh_keyfile
+                )
+                sys.exit(1)
+            except PermissionError:
```
infinitewarp, line 83: I would suggest another exception handler for `UnicodeDecodeError` since we _assume_ the file has readable text but it may not actually.

```suggestion
            except (PermissionError, UnicodeDecodeError):
```
…or maybe a new block with a more specific message to indicate the file having an invalid format.

I discovered this by funneling garbage random bytes to stdin:

```bash
cat /dev/random| head -c 512 | uv run qpc cred add --type network --name $(uuidgen) --username $(uuidgen) --sshkeyfile -
```
```
  File "/Users/brasmith/projects/redhat/qpc/.venv/bin/qpc", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/brasmith/projects/redhat/qpc/qpc/__main__.py", line 11, in main
    CLI().main()
    ~~~~~~~~~~^^
  File "/Users/brasmith/projects/redhat/qpc/qpc/cli.py", line 226, in main
    action.main(self.args)
    ~~~~~~~~~~~^^^^^^^^^^^
  File "/Users/brasmith/projects/redhat/qpc/qpc/clicommand.py", line 83, in main
    self._do_command()
    ~~~~~~~~~~~~~~~~^^
  File "/Users/brasmith/projects/redhat/qpc/qpc/clicommand.py", line 56, in _do_command
    self._build_data()
    ~~~~~~~~~~~~~~~~^^
  File "/Users/brasmith/projects/redhat/qpc/qpc/cred/add.py", line 113, in _build_data
    self.req_payload = build_credential_payload(self.args, self.args.type)
                       ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/brasmith/projects/redhat/qpc/qpc/cred/utils.py", line 125, in build_credential_payload
    req_payload = get_password(args, req_payload, add_none)
  File "/Users/brasmith/projects/redhat/qpc/qpc/cred/utils.py", line 38, in get_password
    req_payload = _get_ssh_key_value(args, req_payload, add_none)
  File "/Users/brasmith/projects/redhat/qpc/qpc/cred/utils.py", line 73, in _get_ssh_key_value
    pass_prompt = "".join(sys.stdin.readlines())
                          ~~~~~~~~~~~~~~~~~~~^^
  File "<frozen codecs>", line 325, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc0 in position 0: invalid start byte
```

infinitewarp (changes requested):
This generally looks good! I just added two questions about possibly unnecessary code and one question about handling an error case.

qpc/cred/utils.py
mirekdlugosz: Good catch.

I think I would draw the line at binary files. As far as I can tell, there is no case where OpenSSH would use private key that is "binary" (i.e. not "text"). So if we can't read and decode file in text mode, it's OK the reject it.

Otherwise, I would leave it up to server. In fact, Ansible runner is our ultimate source of truth about correctness of input data - the only downside is that Ansible runner comes into picture pretty late, prolonging the feedback loop.

abellotti: 👍 bin/text check only. Content validation on server side so the check is there for all users, cli/ui/api.

qpc/cred/add.py
abellotti: just telling the library to use SSH_KEYFILE for display/usage purposes instead of the default dest.

infinitewarp (approved):

qpc/cred/utils.py
abellotti: binary check added (for both tty and stdin) and added tests for both.

PR #377, opened 2025-07-01
feat(source): add proxy support

nicolearagao:
This PR adds support for the proxy_url field in the Add/Edit Source for all non-network sources. It relies on backend proxy validation. 

Relates to JIRA: DISCOVERY-433

## Summary by Sourcery

Add support for specifying and handling a proxy_url in the source add and edit CLI commands, including payload construction and validation via backend.

New Features:
- Add optional --proxy-url argument to source add and edit commands
- Include proxy_url field in the source payload sent to the API

Enhancements:
- Default proxy_url to null when not provided
- Update source tests to cover valid and invalid proxy_url scenarios

### DIFF
--- a/qpc/tests/source/test_source_edit.py
+++ b/qpc/tests/source/test_source_edit.py
@@ -421,3 +421,74 @@ class TestSourceEditCli:
                 CLI().main()
                 expected_message = messages.SOURCE_UPDATED % "source1"
                 assert expected_message in caplog.text
+
+    def test_edit_proxy_url(self, caplog):
+        """Testing the edit source command with proxy_url argument."""
+        url_get_cred = get_server_location() + CREDENTIAL_URI + "?name=credential1"
+        url_get_source = get_server_location() + SOURCE_URI + "?name=source1"
+        url_patch = get_server_location() + SOURCE_URI + "1/"
+        cred_results = [{"id": 1, "name": "credential1", "username": "root"}]
+        cred_data = {"count": 1, "results": cred_results}
+        results = [
+            {
+                "id": 1,
+                "name": "source1",
+                "hosts": ["1.2.3.4"],
+                "credentials": [{"id": 2, "name": "cred2"}],
+            }
+        ]
+        source_data = {"count": 1, "results": results}
+        with requests_mock.Mocker() as mocker:
+            mocker.get(url_get_source, status_code=200, json=source_data)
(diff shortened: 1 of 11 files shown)


ruda (approved):

qpc/tests/source/test_source_edit.py
```diff
@@ -421,3 +421,74 @@ def test_edit_source_port_0(self, caplog):
...
+        with requests_mock.Mocker() as mocker:
+            mocker.get(url_get_source, status_code=200, json=source_data)
+            mocker.get(url_get_cred, status_code=200, json=cred_data)
+            mocker.patch(url_patch, status_code=200)
+
+            args = Namespace(
+                name="source1",
+                hosts=["1.2.3.4"],
+                cred=["credential1"],
+                proxy_url=None,
```
mirekdlugosz, line 486: That doesn't actually clear the proxy, does it?

In `qpc/source/utils.py`  we have this line:

```
if hasattr(args, "proxy_url") and args.proxy_url:
```

So `proxy_url` won't be added to payload unless `proxy_url` is truthy. You can't assign it None or empty string.

Using command line, I don't see a way to clear a proxy. If I have `--proxy-url` flag, it needs a value. And value can't be `''` or `None`. If I skip `--proxy-url`, then server updates other fields as requested, but proxy stays intact.

I think we have the same problem with other optional fields here, like `ssl_cert_verify` (once you assign it value, you can change it, but you can't tell server to return it to default). So I'm not particularly worried. But I feel this test is a bit deceptive - maybe we have mocked out too much here?

nicolearagao: hey, nice catch! Thanks for pointing that out. The current logic in utils.py skips proxy_url entirely if it's falsy, so we're not actually clearing the proxy, just omitting the field. We patch the server with mocker.patch(url_patch, status_code=200), and since we don’t inspect the result on the server, only the request payload, the mock doesn’t simulate the server retaining the original value, which it would in real life. The source data from url_get_source already includes a proxy, but the patch request omits proxy_url, so it stays unchanged. I changed the test name to reflect what this test is actually doing. If we ever want to support actually clearing the proxy, we’d probably need to add a new flag, since the current behavior doesn’t support that directly.

mirekdlugosz (approved):
Thanks for your work on addressing my comments!

PR #379, opened 2025-07-14
refactor(source): switch bulk delete to use v2

nicolearagao:
Replaces usage of the v1 bulk delete API with the v2 version to align
 with updated backend functionality. No UI behavior changes expected.

 Relates to JIRA: DISCOVERY-1006


mirekdlugosz (approved):
Jenkins fails because some tests execute `qpc source clear --all`, but it runs against server without https://github.com/quipucords/quipucords/pull/2998 .

Standalone job #340 is set up to use this branch, server PR #2998 and UI PR #642. There, it all passes.

We can merge server PR and re-run Jenkins here before merging.

PR #385, opened 2025-09-16
feat: add --table option argument to scan list command

ruda:
Add a "--table" option to the "scan list" command to print a simplified table of scan results (human-readable output).

Example:
```
$ qpc scan list --table
 scan_id | scan_name           | report_id | status    
---------+---------------------+-----------+-----------
 3       | xyz                 | None      | None      
 2       | my-new-scan-with-mw | 2         | completed 
 1       | my-scan             | 1         | completed 
```

Relates to JIRA: DISCOVERY-1084

## Summary by Sourcery

Add a "--table" option to the scan list command for human-readable table output, leveraging a new tabular_format utility to render JSON lists as tables and updating tests accordingly.

New Features:
- Add --table flag to scan list command to enable tabular output of scan results.

Enhancements:
- Introduce tabular_format and _json_data_deep_get utilities for rendering JSON data in table form.

Tests:
- Update scan list tests to include output_table flag in test arguments.

### DIFF
--- a/qpc/scan/list.py
+++ b/qpc/scan/list.py
@@ -9,7 +9,7 @@ from qpc import messages, scan
 from qpc.clicommand import CliCommand
 from qpc.request import GET
 from qpc.translation import _
-from qpc.utils import pretty_format
+from qpc.utils import pretty_format, tabular_format
 
 logger = getLogger(__name__)
 
@@ -35,6 +35,12 @@ class ScanListCommand(CliCommand):
             [codes.ok],
         )
         self.parser.add_argument(
+            "--table",
+            dest="output_table",
+            action="store_true",
+            help="Provide tabular formatting of output data (human-readable format).",
+        )
+        self.parser.add_argument(
             "--type",
             dest="type",
             choices=[scan.SCAN_TYPE_CONNECT, scan.SCAN_TYPE_INSPECT],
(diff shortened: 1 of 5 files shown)


infinitewarp (changes requested):
I checked ubi8 and ubi9 for this `prettytable` package, and it looks like Red Hat does not distribute an RPM for modern Python versions. In ubi8, I found a package for Python 3.6, and in uib9, I found a package for Python 3.9.

Also, the upstream project is not GPLv3 licensed. So, we can't easily just vendor/copy over the code as an alternate solution here. I would probably advise against vendoring anyway because it's thousands of lines of code over several files.

We may need to drop using that package as the solution to this feature.

abellotti:
Thanks @ruda for the PR. This looks good 👍 

It's good to have the default attributes for the table for simplifying the report upload user experience.

Though with minimal change, nothing stops us for having a more general purpose --table option for any of the json outputs and not just scan list.  

Thinking of column=attr pairs:

```
qpc scan list --table  "scan_id=id, scan_name=name, report_id=most_recent.report_id, status=most_recent.status"
```

Maybe a future enhancement 😄 

Thank you.

infinitewarp (changes requested):
This looks good overall, but I have a request. It looks like we don't explicitly test `_json_data_deep_get` behaviors, but I think it's complicated enough (and hard to read at a glance) to warrant a specific set of tests to verify that it correctly handles different types and both successfully and unsuccessfully finding nested results with the `.` key syntax. It looks like it works fine today, but unit tests around that function will help prevent regressions if we ever have to change it in the future.

infinitewarp (approved):
Thank you for adding the parametrized unit test to cover happy and edge cases!

LGTM 👍

PR #407, opened 2026-03-05
refactor: bump QPC_MIN_SERVER_VERSION to 2.5.0

nicolearagao:
Remove per-subcommand min_server_version overrides from report deployments, details, download, and insights commands. Mixing old and new server/client is not supported, so all commands should use the single global QPC_MIN_SERVER_VERSION. Bump it from 2.4.0 to 2.5.0 to reflect the new APIs required by report show and list.

Relates to JIRA: DISCOVERY-1272

## Summary by Sourcery

Align CLI minimum server version requirements with a single global value and update it to 2.5.0.

Enhancements:
- Centralize min server version validation on the global QPC_MIN_SERVER_VERSION constant instead of per-report subcommand overrides.
- Update the global CLI minimum supported server version from 2.4.0 to 2.5.0 to reflect new server API requirements.

### DIFF
--- a/qpc/clicommand.py
+++ b/qpc/clicommand.py
@@ -25,8 +25,7 @@ class CliCommand:
         self.req_headers = None
         self.response = None
 
-        # If you add or change API, you must update these versions
-        # this includes self.min_server_version
+        # Minimum server version required by the CLI
         self.min_server_version = QPC_MIN_SERVER_VERSION
 
     def _validate_args(self):
--- a/qpc/report/deployments.py
+++ b/qpc/report/deployments.py
@@ -74,7 +74,6 @@ class ReportDeploymentsCommand(CliCommand):
             help=_(messages.REPORT_PATH_HELP),
         )
         self.report_id = None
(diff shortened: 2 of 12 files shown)


mirekdlugosz (commented):
I don't understand what is happening in unit tests, but I *think* you should also remove all `headers={"X-Server-Version": VERSION}` from mocks ([example](https://github.com/quipucords/qpc/blob/2bea4fd064e1b2588623dd7c12e9e610dfd0c838/qpc/tests/report/test_report_list.py#L61)).

I see only tests in `qpc/tests/report/` set that, and it so happens that only tests in `qpc/tests/report/` are failing CI. Although I have not checked that all failing tests have that `headers` set.

Also, personally I would not include `QPC_MIN_SERVER_VERSION` bump here. We routinely do that as part of new version release. Yes, I know jira ticket explicitly says that. I still think jira ticket mixes two different things and forces you to create a PR that can't be merged due to CI failures.

abellotti (approved):

PR #428, opened 2026-08-13
refactor: set QPC_VAR_PROGRAM_NAME to a constant

FabriciaDinizRH:
QPC_VAR_PROGRAM_NAME used to be an environment variable, but it was not being used by anyone.
To simplify building downstream, it's being changed to a constant as a part of [DISCOVERY-1244](https://redhat.atlassian.net/browse/DISCOVERY-1244).

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized the program name to `qpc` across command-line, container, and packaged installations.
  * Prevented environment settings from overriding the displayed program name.
  * Ensured RPM packaging uses the correct program name during installation.

### DIFF
--- a/qpc.spec
+++ b/qpc.spec
@@ -43,7 +43,7 @@ qpc is the command-line client interface for the quipucords server.
 
 %build
 sed -i \
-  -e 's/^QPC_VAR_PROGRAM_NAME = os.environ.get("QPC_VAR_PROGRAM_NAME", "qpc")/QPC_VAR_PROGRAM_NAME = os.environ.get("QPC_VAR_PROGRAM_NAME", "%{binname}")/' \
+  -e 's/^QPC_VAR_PROGRAM_NAME = "qpc"/QPC_VAR_PROGRAM_NAME = "%{binname}"/' \
   %{_builddir}/qpc-%{version}/qpc/release.py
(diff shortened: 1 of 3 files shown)


qpc.spec
```diff
@@ -70,6 +70,9 @@ sed \
 %{python3_sitelib}/qpc-*.egg-info/
 
 %changelog
+* Wed Aug 12 2026 Fabricia Diniz <agomesdi@redhat.com> - 0:2.8.0-1
+- Update sed to read QPC_VAR_PROGRAM_NAME as a constant.
+
```
infinitewarp: A few thoughts:

1. This would be only a `.z`-level change since it's a very minor bugfix change from the user's perspective. So, `2.7.1` instead of `2.8.0`
2. We keep our `x.y` versions in sync across all our projects. So bumping this to `2.8.0` would mean not releasing `qpc` until all the other projects are _also_ ready to bump to `2.8.0`.
3. We probably shouldn't bump the version in the changelog without also bumping the *actual* version up at the top of the spec file, which we're not doing here. It's typically the release engineer's responsibility to bump versions and update changelogs if necessary.

```suggestion
```

infinitewarp (approved):

